PR #9, opened 2015-01-15
Make CSS easier to read

starcalibre:
Added a few things to make the CSS a little easier to read:

* Add comments to better separate between sections
* Add comments to explain some of the less obvious parts
* More consistency with colour selections and formatting

Also a small change to the scatterplot function..

* Use CSS to style scatterplot points where possible, instead of using the D3 selector.

@jni - Not sure if you'll be notified of this so pinging you just in case :)

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -1,4 +1,5 @@
-/* Main Page CSS */
+/* MAIN PAGE CSS */
+
 .container {
     width: 95%;
 }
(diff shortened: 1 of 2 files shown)


public/css/style.css
```diff
@@ -1,4 +1,5 @@
-/* Main Page CSS */
+/* MAIN PAGE CSS */
```
jni: WHY ARE YOU YELLING

public/css/style.css
```diff
@@ -17,26 +19,39 @@
...
+    display: inline-block;
+    float: none;
+    text-align: left;
+    margin-right: -4px;
 }
 
+/* Centered rows for grid display in Nebula */
 .row-centered {
-    text-align:center;
+    text-align :center;
```
jni: Why put a leading instead of trailing space? Guessing this is a typo... =)

public/css/style.css
```diff
@@ -17,26 +19,39 @@
...
 .plotbox {
     text-align: center;
 }
 
+/* GENERAL PLOT CSS */
+
+.axis path, .axis line {
+    fill: none;
+    stroke: black;
+    shape-rendering: crispEdges;
```
jni: Haha this is great value. I think we might want to go with superCrisp though. =P

jni:
Very clean! Just one very minor comment. =)

public/css/style.css
starcalibre: That's a typo :P Thanks

starcalibre:
@jni this should be good to merge :)

jni:
So it is. =)

Issue #7, opened 2015-01-12
Mongo schema


starcalibre:
# Mongo

    var Screen = {
        _id: String, // short screen name < 8 char
        screen_name: String, // full screen name
        screen_desc: String,
        number_samples : Number
        features : Array // Array of features used in the screen
    }

Screen has a one-to-many relationship to Sample.

    var Sample = {
            _id: String, // name-plate-well
            screen: String, // Screen _id
            gene_name: String, // or gene_nane
            control_pos: Boolean, // true if sample is positive control
            control_neg: Boolean, // true if sample is negative control
            feature_vector: Array, // unstandardised features
            featute_vector_std: Array, // standardised features
            neighbours: Array, of _id // nearest neighbours (reference to samples)
            column: String, // '01', '02', '03', ...
            row: String, // 'A', 'B', 'C', ...
            pca_vector: Array,
            image: ObjectID, // Image ID
            cluster_member: Array, // array of cluster memberships for clustering scheme k=2..20
        }

    var Image = {
            sample_id: String, // name-plate-well ENSURE A UNIQUE INDEX IS CREATED HERE!
            image_full: String, // BinData of image
            image_thumb: String // BinData of image (thumbnail)
        }
Screen has a one to one relationship with Images.

Documents will be stored in the ``microscopium`` database, Screen documents in ``screens`` collection and Sample documents in ``samples`` collection. Image documents in ``images`` collection.

starcalibre:
This information is now available in the readme that added in #34

PR #11, opened 2015-01-19
Switch between different screen datasets

starcalibre:
* Added a line to the scatterplot function to prevent it from drawing twice
* Changed the feature line plot to display the values of the standardised feature vectors
* Updated the routes and dropdown selector to allow the user to switch between multiple datasets

Note: The routes, AJAX requests and D3 functions assume the MongoDB models detailed in #7 are being used.

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -9,14 +9,25 @@ module.exports = function(app) {
         res.sendfile('index.html');
     });
 
-    // get all screen documents
-    app.get('/api/screens', function(req, res) {
+    // get all screen ids
+    app.get('/api/screen_ids', function(req, res) {
         Screen
-          .find({})
-          .exec(function(err, data) {
-            if (err) res.send(err);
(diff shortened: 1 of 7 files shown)


app/routes.js
```diff
@@ -9,14 +9,25 @@ module.exports = function(app) {
...
+    app.get('/api/screen_ids', function(req, res) {
         Screen
-          .find({})
-          .exec(function(err, data) {
-            if (err) res.send(err);
-            res.json(data);
-          });
+            .find({})
+            .select('_id')
+            .exec(function(err, data) {
```
jni: define a separate function for this since it is repeated at least thrice. (love that word.)

jni:
Looks very clean! Just one change required pre-merging. =)

jni:
Awesome! This illustrates how code review is a great idea even from a total n00b! =D

PR #12, opened 2015-01-22
Single screen nebula

starcalibre:
Fairly big PR, we could potentially go through this one tomorrow as part of the next UI walkthrough..

* Combine the neighbour image gallery and the PCA plot
* Add the PCA plot with slider to change the cluster labels
* Fix indentations in ``viz-scatterplot.js`` and ``viz-neighbours.js``

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -28,6 +28,13 @@
     height: 150px;
 }
 
+.selected-image-display {
+    margin: 1px 1px -3px 1px;
+    padding: 0 0 0 0;
+    width: 300px;
+    height: 300px;
+}
+
 /* Centered columns for grid display in Nebula */
 .col-centered {
     display: inline-block;
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -80,7 +80,16 @@
     stroke-width: 2px;
 }
 
-/* SCATTERPLOT CSS */
+/* NEIGHBOUR PLOT CSS */
+
+.gallery {
+    height: 500px;
+    overflow: scroll;
+}
+
+.pcaplot {
+    height: 500px;
+}
 
 .scatterpt {
     fill: steelblue;
@@ -92,7 +101,7 @@
 }
 
 .neighbourpt {
-    fill: yellow;
+    fill: #ffb323;
 }
 
 /* TOOLTIP CSS */
--- a/public/index.html
+++ b/public/index.html
@@ -53,6 +53,11 @@
             </a>
         </li>
         <li>
+            <a href="#clusterplot" data-toggle="tab">
+                Cluster Cosmologist
+            </a>
+        </li>
+        <li>
             <a href="#importexport" data-toggle="tab">
                 Export Data
             </a>
@@ -241,6 +246,20 @@
                 </div>
             </div>
         </div>
+        <!-- CLUSTER COSMOLOGIST -->
+        <div class="tab-pane fade" id ="clusterplot">
+            <div class="row">
+                <div class="row">
+                    <div class="col-lg-4 plotbox">
+                        <label for="cluster-slider">Number Clusters:</label>
+                        <input id="cluster-slider" type="range" min="2" max="20" value="2" step="1"/>
+                        <span id="cluster-number">2</span>
+                    </div>
+                    <div id="scattercluster" class="col-lg-8 plotbox">
+                    </div>
+                </div>
+            </div>
+        </div>
         <!-- IMPORT EXPORT TAB -->
         <div class="tab-pane fade" id="importexport">
             <div class="row">
@@ -256,5 +275,6 @@
 <script src="js/viz-scatterplot.js"></script>
 <script src="js/viz-lineplot.js"></script>
 <script src="js/viz-neighbours.js"></script>
+<script src="js/viz-scattercluster.js"></script>
 <script src="js/script.js"></script>
 </html>
--- a/public/index.html
+++ b/public/index.html
@@ -48,12 +48,7 @@
             </a>
         </li>
         <li>
-            <a href="#clusterexplorer" data-toggle="tab">
-                PCA Commander
-            </a>
-        </li>
-        <li>
-            <a href="#imageexplorer" data-toggle="tab">
+            <a href="#neighbournebula" data-toggle="tab">
                 Neighbour Nebula
             </a>
         </li>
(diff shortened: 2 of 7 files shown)


jni:
Yeah let's have a look in detail tomorrow. In the future though keep indent fixes completely separate, because they obliterate the diff!

public/css/style.css
```diff
@@ -28,6 +28,13 @@
     height: 150px;
 }
 
+.selected-image-display {
+    margin: 1px 1px -3px 1px;
+    padding: 0 0 0 0;
+    width: 300px;
+    height: 300px;
+}
+
```
jni: Just a comment for my own enlightenment: is this a matter of trial and error? Or is there a method to the madness? e.g. a commonly-cited UI design manual?

jni: And actually, maybe we want to use horizontal percentages instead of pixel values? This way we won't have issues with the divs stacking on top of each other when the window is resized... Thoughts?

public/index.html
```diff
@@ -93,148 +93,169 @@
...
-                            <img id="nebula-18" class="img-responsive" src="" alt="">
-                        </a>
-                    </div>
-                    <div class="col-lg-3 col-centered thumb">
-                        <a class="thumbnail" href="#">
-                            <img id="nebula-5" class="img-responsive" src="" alt="">
-                        </a>
+                <div class="col-lg-5 gallery">
+                    <div class="row-centered">
+                        <div class="col-lg-3 col-centered selected-image-display">
```
jni: Tell me a bit more about these CSS classes? ie what's the number in `col-lg-x`? Is there different sizes within the class large? And if so, doesn't it make more sense to have `col-size-x` for all kinds of columns? Also, I'd rather be more explicit in these names (lg->large).

public/index.html
```diff
@@ -93,148 +93,169 @@
...
+                            </a>
+                        </div>
+                        <div class="col-lg-3 col-centered thumb">
+                            <a class="thumbnail" href="#">
+                                <img id="nebula-8" class="img-responsive" src="" alt="">
+                            </a>
+                        </div>
+                        <div class="col-lg-3 col-centered thumb">
+                            <a class="thumbnail" href="#">
+                                <img id="nebula-9" class="img-responsive" src="" alt="">
```
jni: Whooaaaa... Isn't there a way to generate this programmatically? (The nebula-x's, that is.)

public/index.html
starcalibre: Those are classes are built into Bootstrap that make the task of dividing your screen into x number of columns much easier. It works by splitting the screen up into 12 equal size columns, then you specify how wide you'd like your columns to be.. eg two ``col-lg-6`` will give you two equal columns. The number needs to add up to 12.

See these examples on the Bootstrap website: http://getbootstrap.com/examples/grid/

jni: Oh well if the classes are built-in, screw all of my ideas for renaming them LOL

jni: One option for noobs such as myself would be to put a comment on first use of a non-obvious bootstrap class in a file, that explains the class and its purpose. For example, there would have been no way for me to know that the classes needed to add to 12. Other class names, such as "selected-image-display", are nice and self-explanatory. =)

Could you add such a comment above somewhere?

public/css/style.css
starcalibre: Working with CSS tends to be a combination of knowledge, trial and error and stack overflow.. :P There's plenty of great UI texts out there, but there's no authority on CSS. 

Yes, absolutely -- responsiveness is something to think about. On my 24" monitor at home, there's lots of wasted screen real estate when I ruin the UI. Once we know exactly how our pages will be laid out, I'd like to work in % too. Once we switch over to node-web-kit, it's possible to specify a minimum resolution for the window. So we could work on that minimum resolution, then just ensure it resizes well to larger screens.

starcalibre:
Yeah that's an easy for-loop that calls jQuery -- we'll probably need to do it that way once we change the 'view neighbours' option to view neighbours based on a distance threshold, as the number of neighbours will vary.

public/css/style.css
jni: Ok, we can leave this to later in the development process. Could you add an issue to change it to % to make sure we track this?

My preferred min res is 840, because it allows people to have two browser windows side-by-side if they have 1680 horizontal resolution. I realise it might be a bit low — what do you think? Anyway move this discussion to the issue itself. =)

public/index.html
```diff
@@ -93,148 +93,169 @@
...
-                        </a>
+            </div>
+        </div>
+        <!-- CLUSTER COSMOLOGIST -->
+        <div class="tab-pane fade" id ="clusterplot">
+            <div class="row">
+                <div class="row">
+                    <div class="col-lg-4 plotbox">
+                        <label for="cluster-slider">Number Clusters:</label>
+                        <input id="cluster-slider" type="range" min="2" max="20" value="2" step="1"/>
```
jni: Put this in the intermediate value, ie 10? 2 looks really low for most cases, and makes the display look a bit broken.

public/js/viz-neighbours.js
```diff
@@ -1,30 +1,47 @@
...
-      for(var i = 0; i < 25; i++) {
-        var nebula_selector = '#nebula-' + i;
-        $(nebula_selector).attr('src', 'data:image/jpg;base64,' + json[i]['image_thumb']);
-        $(nebula_selector).attr('title', json[i]['sample_name sample_id']);
-      }
+function setImages(selectedImage, neighbourImages) {
+    $('#nebula-0').attr('src', 'data:image/jpg;base64,' + selectedImage['image_full']);
+    $('#nebula-0').attr('title', selectedImage['sample_id']);
+
+    for(var i = 0; i < 24; i++) {
```
jni: Don't hardcode the 24... At least make it a global variable, or perhaps something to be read from a config file? Something that will allow us to mess with this easily. Magic numbers are bad. =)

public/js/viz-scattercluster.js
```diff
@@ -0,0 +1,115 @@
+// TODO rewrite scatterPlot and controls as an object
+var renderScatterCluster = function(sampleData, k) {
+
+    // delete any scatterplots already plotted
+    d3.select('#scattercluster > svg').remove();
```
jni: Haha yeah this is crazy! =D Can't one recolour existing dots? If so but you don't want to mess with it in this PR, please add an issue and respond here so I know it's "resolved". =) But we are accumulating a worrying amount of technical debt...

jni:
Can you add the requested changes to this PR?

- [ ] comment re: bootstrap CSS classes
- [ ] remove magic number of neighbours in for loop
- [ ] cluster slider default value (come to think of it, make the min and max clusters also easily changeable, without magic numbers)
- [ ] (optional, with new issue added) programmatically add the neighbour display HTML
- [ ] (optional, with new issue added) recolour dots rather than redrawing from scratch.

starcalibre:
Whoah, I can check those tickboxes off? That's awesome! :D

I can do all but the last one this afternoon. If I'm going to re-write any D3 plots, I'd rather also be re-writing them as React components.. see issue #17

jni:
Haha yeah, it's a Github Flavored Markdown (GFM) feature. You use `- [ ]` for an empty tick box, `- [x]` for a pre-ticked one. Both are editable. =D

Awesome Re the progress! Feel free to tick off that last point since you have already created the corresponding issue! =)

starcalibre:
Issue #18 added for the second-last change, the rest are all done! :D

jni:
All good! =)

PR #20, opened 2015-01-29
Lint JS files to keep JSHint happy

starcalibre:
Now we've got grunt and JSHint running on master, I've gone through and corrected all JSHint warnings..

* These were primarily warnings to use dot notation instead of bracket notation when accessing object properties, ie use ``object.property`` instead of ``object['property']``. I personally prefer dot notation, and it's the preferred style by [Airbnb](https://github.com/airbnb/javascript#properties) and [Crockford](http://javascript.crockford.com/survey.html)
* There were also a couple of missing semi-colons and unused global variables

### DIFF
--- a/public/js/script.js
+++ b/public/js/script.js
@@ -1,11 +1,9 @@
-var featureNames = [];
-
 // parse feature info
 var getFeatureRow = function(json, feature) {
     var n_features = json.length;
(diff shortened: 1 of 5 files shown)


jni:
That was straightforward enough!!!

PR #21, opened 2015-01-29
Amend gruntfile to watch all js files

starcalibre:
Pretty straightforward.. Grunt wasn't watching all the js files.


jni:
Ah yes, `**` doesn't work on all flavours of *nix. That lazy grunt... =P

PR #22, opened 2015-01-29
Gene tooltips

starcalibre:
* Tooltips now show the gene associated with the sample (where available)
* Fixed a bug that was preventing the tooltips from showing the updated cluster membership ID after the number of clusters was changed on the slider.


jni:
Straightforward enough! Thanks!!! =)

PR #23, opened 2015-02-09
Update initial DB query on pageload

starcalibre:
The first of a couple of PRs that will add the filtering function to the UI.. this one is fairly simple!

* Reduce the number of fields returned by Mongo when a screen is selected. This greatly reduces loading times.
* Fix a bug that was preventing the title from showing in the lineplot
* Fix a bug that was causing the cluster scatterplot to sometimes render to the wrong div
* Redirect users to the summary screen whenever the active screen is changed

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -35,6 +35,8 @@ module.exports = function(app) {
     app.get('/api/samples/:screen', function(req, res) {
         Sample
         .find({ 'screen': req.params.screen })
+        .select('_id row col gene_name feature_vector_std ' +
+                'pca cluster_member neighbours')
         .exec(resHandler(res));
     });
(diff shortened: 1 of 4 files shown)


app/routes.js
```diff
@@ -35,6 +35,8 @@ module.exports = function(app) {
     app.get('/api/samples/:screen', function(req, res) {
         Sample
         .find({ 'screen': req.params.screen })
+        .select('_id row column gene_name feature_vector_std ' +
+                'pca cluster_member neighbours')
```
jni: Very nice!

jni:
Wow, that was easy! =) Very nice fixes!

PR #24, opened 2015-02-10
Add slidebar and refactor index.html

starcalibre:
Getting close to adding the filter functionality now!

* Add the slidebars plugin, and rewrite structure of the index.html and various other jQuery methods so everything still functions well with the slidebar.
* Remove the tabbed navigation, and move the view control to the navbar. This frees up screen real estate and makes everything look a bit cleaner.
* Fix a small bug in the cluster scatterplot that causing an empty SVG to be added to the DOM when the number of clusters was changed.

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -119,6 +119,7 @@
     background: rgba(0, 0, 0, 0.8);
     color: white;
     border-radius: 2px;
+    z-index: 999999;
 }
 
 /* Creates a small triangle extender for the tooltip */
--- a/bower.json
+++ b/bower.json
@@ -7,6 +7,7 @@
         "d3": "latest",
         "d3-tip": "latest",
         "jquery": "latest",
-        "spin.js": "latest"
+        "spin.js": "latest",
+        "slidebars": "latest"
     }
 }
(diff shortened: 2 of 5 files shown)


public/css/style.css
```diff
@@ -119,6 +119,7 @@
     background: rgba(0, 0, 0, 0.8);
     color: white;
     border-radius: 2px;
+    z-index: 999999;
```
jni: Is this assuming that we won't have > 1M points?

jni:
Hmm, tried to play with this... `node server.js`... Navigate to localhost:8080... I get a page greeting me with a friendly "Yo", as well as an empty drawer for "Select screen"... Suggestions?

starcalibre:
Run ``bower install && npm install`` to make sure you've got all the
dependencies. If it still doesn't work, can you check for any errors in
your browsers console?

public/css/style.css
starcalibre: It's just an arbitrarily large number. Though I assume we won't be drawing
1M SVGs! Datasets that large would be pointless viewed on a scatterplot
anyway. That's something where 2D binning would be useful :)

jni:
That is of course what was wrong! Looks *great!* And I need to get in the habit of checking the console, that would probably have told me. imho browsers should have a little "hey all sorts of shit is happening in the JS console" button that lights up when things go wrong. =)

PR #25, opened 2015-02-17
Add (non-functioning) filter

starcalibre:
Almost finished!!

This PR adds a few AJAX calls to the screen selection function that pulls unique values for plates, rows, columns and treatments (currently stilled called ``gene_names``) and populates the contents of the filter in the slidebar.

It doesn't do any filtering yet, as I need to update all the plotting functions to have update methods (so we can update the plots without having to redraw them completely).

PS - I've just noticed the MYORES dataset does not have a ``plate`` field, so this will currently error if you try and load it. Works fine for the LNC datasets though. I'll update the DB to include the plate data in MYORES and give you a copy of that JSON today. :)


app/routes.js
```diff
@@ -24,6 +24,16 @@ module.exports = function(app) {
             .exec(resHandler(res));
     });
 
+    // find unique values for given screen and field
+    app.get('/api/unique/:screen/:field', function(req, res) {
+        Sample
+            .find({'screen': req.params.screen})
+            .distinct(req.params.field, function(err, data) {
```
jni: erm, use resHandler? =)

jni: Oh. I see the sort. Does js have optional args? ie could you add a "sort=False" arg to resHandler, then call `resHandler(res, true)` here?

public/css/style.css
```diff
@@ -17,6 +18,44 @@
...
+.nav-stacked li a:focus {
+    background-color: #000000;
+}
+
+/* Force sidebar button to use 100% of sidebar width */
+.sidebar-container {
+    width: 100%;
+}
+
+/* Override speed of collpase menu */
```
jni: collpase -> collapse

app/routes.js
starcalibre: Long answer:
JS actually doesn't check the number of arguments passed.. normally optional arguments are handled something like this.

```
function excellentFunction(optionalArgument) {
    if(optionalArgument === undefined) {
        optionalArgument = 'defaultValue';
    }
    return optionalArgument;
}
```

It's one of the annoyances of ES5 (the current version of JS), though optional arguments are coming in ES6 which is pretty exciting.

Short answer:
Yeah that can be done :)

starcalibre: Oh and just so you know why I've done the sorting myself.. Mongo annoyingly doesn't let you sort output in a distinct query. http://mongoosejs.com/docs/api.html#query_Query-sort

public/index.html
```diff
@@ -262,11 +262,51 @@
...
+                        <div class="collapse container filter-content" id="col-filter">
+                            <div class="row">
+                                <div class="col-md-4 filter-items" id="col-1"></div>
+                                <div class="col-md-4 filter-items" id="col-2"></div>
+                                <div class="col-md-4 filter-items" id="col-3"></div>
+                            </div>
+                        </div>
+                    </li>
+                    <li>
+                        <a href="#gene-filter" data-toggle="collapse">Gene</a>
```
jni: Let's move this to the top of the filter list. Am I supposed to be able to click on this box and type? Cos it's not working for me...

jni:
As a reminder, I don't get notifications when you push here, so ping me when the next push happens!

starcalibre:
@jni -- the functionality is all there now. If you run this branch locally, you'll need to run ``bower install`` because I needed to add jQuery plugin to help parse the form data. 

If you select some filter parameters and hit the "Apply" button, in the console you'll see the object that gets produced by the form. That's what will be used to update the data!

starcalibre:
I should probably clarify the filter works in terms of being able to interact with it, but it won't alter the data being displayed.. yet ;)

jni:
@starcalibre LOL *should -> should have. =P

btw, I was thinking, re: data display, rather than making unfiltered genes disappear, make them have a really low alpha? That helps keep the filtered samples in context.

jni:
Cool, works for me, after stupidly battling with the install. I didn't have coffee this morning, ok??? =P (Making one now, hopefully before I collapse from mental shutdown.) Anyway, imho the Gene filter box should be expanded by default, but otherwise works great!

And if it's close to trivial, one nice thing would be for the selection to drop down to the next gene, rather than disappearing. e.g. search for p53 (3 genes appear). Click the first one, and it drops down, and the selection disappears. If I want to add the next one, I have to go click on it, then go back to the arrow, etc. It'd be nice to be able to click the arrow three times. (But this is a **minor nitpick** and if there's *any* difficulty at all it should be put very far on the backburner!)

starcalibre:
@jni - Done, great suggestions! Adding genes feels much better now. I don't really consider suggestions like that a nitpick, minor annoyances can compound into a frustrating UI.

jni:
And that was *before* my coffee! =P

Very nice!

> minor annoyances can compound into a frustrating UI.

That's very well put and has prompted another idea (which again might be hard to implement): pressing Enter adds the currently selected gene (the top one as you're typing) to the filter, *and* selects the whole text box so that more genes can be searched for instantly, without ever having to touch the mouse.

Also btw, do you prefer sliding the whole window over having a drawer overlay? I just saw the overlay-style UI in Betterment (an investment company) for notifications and it feels much less disruptive.

starcalibre:
Shouldn't be too hard!

What an awesome website! Hopefully investment returns are correlated with a good responsive UI ;)  I can't quite see what you're referring to though? Do you have another example?

starcalibre:
@jni - Let me know what you think. I'm pretty indifferent about what style of slidebar I prefer. I'll probably have more of an opinion once the filter actually works ;P

jni:
Just to formalise what we discussed earlier today, here's what needs to be updated before I merge the PR:

- change "Filter" to a filter icon (the little funnel)
- the icon overlays everything, so that it remains under the cursor when clicked, and users can just click once to show the drawer, and click again to hide it (no need to move the mouse)
- the icon changes colour depending on whether any filters are currently active. I suggest grey for inactive, white for drawer open, red for active.
- the updates should happen live as things are added to the filter, so remove the "apply" button. (but keep reset.)

starcalibre:
@jni These have been done, but with one exception! [This](https://i.imgur.com/Q3cUg29.gif) is an accurate depiction of my experience getting the filter icon to stay white while the drawer is open. I'm leaving it for the moment !

jni:
LOL That was hilarious! I'm merging this but I'll make a few issues (ie this is technical debt) (which sounds bad but each of these is small enough that it probably makes more sense to fix separately in their own time, rather than hold this PR back):

- We do need to figure out how to keep the icon coloured. Actually, there's two colours I want: (1) white while the drawer is open; this is minor because the user can be perceptive and just notice that yes, the drawer is open. =P But also (2) I want it to be red, or badged, or somehow changed when there is an active filter.
- I want the keyboard focus to go straight to the gene box when one clicks on the icon. We can at some point even include a keyboard shortcut for the icon, so the users can work with it without taking their hands off the keyboard.
- I assume the blank line represents samples with no gene name. That should be replaced by a suitable explanatory string, e.g. "[no gene info]".

Issue #26, opened 2015-02-17
Ensure Neighbour Nebula display is appropriately resized for less-wide browsers


jni:
If the browser window is resized to be less than ~1280 pixels wide, the images get stacked *under* the PCA plot, which looks hideous and is not functional. Hopefully some kind of rescaling of the PCA plot is possible...

starcalibre:
I think I should rewrite the UI where necessary so the minimum resolution supported is 1024x768 -- 99% of web users have a resolution greater than this ([source](http://www.w3schools.com/browsers/browsers_display.asp)). 

Then I can look adding responsiveness so screen real estate isn't wasted in higher resolutions. It's much easier to think small first then scale everything up to larger screens than the other way around!

jni:
If it's not too much of a challenge, I'd like to have the minimum workable res to be ~720. That allows a side-by-side browser window when snapped to the left and right sides of the screen on a resolution of 1440x900 — typical of 14- and 15-inch laptop monitors.

btw, from your source:
> the numbers below might not be 100% true for the average internet users. The average internet-user might have a lower screen resolution.

;)

To say nothing of biologists! =P

jni:
btw, I know 720 might not be reasonable/workable. The next two side-by-side resolutions are 840 (for 1680 horizontal res) and 960 (for 1920 horizontal res). Anyway I just wanted to make sure these are considerations. They should not drive the design but check it.

PR #34, opened 2015-02-25
Add Browserify and re-write plot functions as objects

starcalibre:
Okay, this is a pretty significant PR.. but I think this takes the quality of the code from 'something hacked together by a noob who has no idea WTF' to something approaching quality. Happy to run through it in person if need be!

* Plots have now been re-written as objects with methods for drawing and updating. This makes the plots much more modular and easier to understand, and when a filter is applied to the data we'll be able to update the bare minimum needed to reflect the new data. The old plots were 'draw once' type functions, so the whole plot needed to be re-drawn if the data changed. Another advantage to this approach is that plots can be re-usable, so we could say draw two lineplots to allow a user to compare feature distributions of two samples at once.

* The 'active' feature selected in the lineplot stays the same even after you select a new sample from the PCA plot.

* All JS files are now Browserified! Now that modules can now be imported with 'require' on the front-end, dependencies are now loaded in NPM instead of Bower where possible. Bower is used for libraries where there's CSS being imported (eg Bootstrap, Font-Awesome, Slidebars) or when we want the library to be global (eg jQuery and jQuery plugins). This means you HAVE to use grunt whenever you run the UI locally to make sure all the JS files get browserified first. Also remember to ``npm install && bower install`` if you run this locally too.

* Lodash has been used whenever possible for array operations. Lodash usually outperforms Javascript for many operations.

Once this is merged, I can add methods to the plots that will update them whenever the Filter is updated.


jni:
Just btw, one of your git user emails is not associated with your github account (see commits above). Might want to fix that. (You can add multiple email accounts to each user account.)

public/js/ClusterPlot.js
```diff
@@ -0,0 +1,166 @@
+var d3 = require('d3');
+require('d3-tip')(d3);
```
jni: Can you explain what this does? (in a comment)

public/js/ClusterPlot.js
```diff
@@ -0,0 +1,166 @@
...
+
+function ClusterPlot(sampleData) {
```
jni: Ok, super-confused here... `ClusterPlot` is a function, an object, both, a constructor? `this` refers to what?

public/js/ClusterPlot.js
```diff
@@ -0,0 +1,166 @@
...
+    this.clusterMin = 2;
+    this.clusterMax = 20;
+    this.clusterMid = Math.round((this.clusterMax+this.clusterMin)/2);
+    this.sampleData = sampleData;
+
+    this.fullWidth = 650;
+    this.fullHeight = 450;
```
jni: I've been trained to despise magic numbers like this... Any way to define them elsewhere? e.g. Why is this not a particular class of plot that has certain stylings as defined here?

public/js/Histogram.js
```diff
@@ -0,0 +1,95 @@
...
+    this.sampleData = sampleData;
+    this.featureNames = featureNames;
+    this.feature = 0;
+
+    this.fullWidth = 500;
+    this.fullHeight = 400;
+
+    this.margin = {top: 20, right: 30, bottom: 20, left: 45};
+    this.width = 500 - this.margin.left - this.margin.right;
+    this.height = 400 - this.margin.top - this.margin.bottom;
```
jni: As above... So many magic numbers... =\

public/js/Histogram.js
```diff
@@ -0,0 +1,95 @@
+var d3 = require('d3');
+var _ = require('lodash');
+_.mixin(require('lodash-deep'));
```
jni: haha, `lodash`, I get it.

Also, whoever came up with it should be shot. =P

public/js/Histogram.js
```diff
@@ -0,0 +1,95 @@
...
+        .domain([d3.min(featureDist), d3.max(featureDist)])
+        .range([0, this.width]);
+
+    var xAxis = d3.svg.axis()
+        .scale(xScale)
+        .orient('bottom');
+
+    var data = d3.layout.histogram()
+        .bins(xScale.ticks(10))
+    (featureDist);
```
jni: Ok this is super-groovy indentation/syntax. Should the call (?) not be aligned with the . from `.bins`?

public/js/Histogram.js
```diff
@@ -0,0 +1,95 @@
...
+        .attr('height', function(d) { return self.height - yScale(d.y); });
+
+    // append x axis
+    svg.append('g')
+        .attr('class', 'x axis')
+        .attr('transform', 'translate(0,' + self.height + ')')
+        .call(xAxis);
+
+    // append y axis
+    svg.append('g')
```
jni: 'g'?

public/js/Lineplot.js
```diff
@@ -0,0 +1,163 @@
...
+    this.activeFeature = 0;
+    this.histogram = histogram;
+
+    this.fullWidth = 600;
+    this.fullHeight = 400;
+
+    this.margin = {top: 20, right: 30, bottom: 20, left: 45};
+
+    this.width = this.fullWidth - this.margin.left - this.margin.right;
+    this.height = this.fullHeight - this.margin.top - this.margin.bottom;
```
jni: Hey! You did it properly here! =P

public/js/NeighbourImages.js
```diff
@@ -0,0 +1,55 @@
...
+            url: '/api/images/',
+            data: {
+                'sample_ids': self.neighbours
+            },
+            async: false,
+            success: function(json) {
+                self.neighbourImages = json;
+            }
+        })).then(function(res, status) {
+            console.log('then!');
```
jni: Can we please log something a bit more descriptive here??? LOL

public/js/NeighbourImages.js
```diff
@@ -0,0 +1,55 @@
...
+
+NeighbourImages.prototype.getImages = function(query_id) {
+    var self = this;
+
+    $.when(
+        $.ajax({
+            url: '/api/sample/neighbours/' + query_id,
+            async: false,
+            success: function(json) {
+                self.neighbours = json[0].neighbours.slice(1, 25);
```
jni: The number of neighbours is another magic number... Is it referenced elsewhere, and if so, a global config file for it would be nice...

public/js/NeighbourPlot.js
```diff
@@ -0,0 +1,168 @@
...
+        .style('text-anchor', 'middle')
+        .text('PC2');
+
+    // default select first point
+    var firstPoint = d3.select('.scatterpt');
+    console.log(firstPoint.data());
+    self.updatePoint(firstPoint, firstPoint.data()[0],  0);
+};
+
+NeighbourPlot.prototype.updatePlot = function() {
```
jni: WAT

public/js/NeighbourPlot.js
```diff
@@ -0,0 +1,168 @@
...
+        .transition()
+        .duration(125);
+
+    // add neighbour class to all neighbours
+    var neighbours = d.neighbours;
+    for(var i = 1 ; i < neighbours.length; i++) {
+        d3.select('[id=' + neighbours[i] + ']')
+            .classed('neighbourpt', true)
+            .transition()
+            .duration(125)
```
jni: Here's a number ripe for codifying: you've typed "125" a total of... Well, not sure, but it feels like 125 times! =P A 'transition_duration' variable that can be set in one place would be really handy here...

jni:
This is as good a time as any to ask you to add running instructions to the README, e.g.:

get the latest changes from git
bower install && npm install
# ensure mongodb is running in the background
node server.js

Navigate to "localhost:8080" on your browser.

I suspect some of those steps might have changed now?

I've made a few comments about magic numbers... Some of them might not be practical to fix, but try to catch as many as possible...

jni:
@starcalibre see also #36. It might be worthwhile tackling it in this PR. (But maybe it's not, your call!)

starcalibre:
Thanks, I'll go through your comments tomorrow :)

I'd rather get this sorted, get the filter working then look at #36 as a separate PR. I think a PR sorting out #36 could also address #26 (ie better responsiveness to varying size monitors).

jni:
Yep, no worries, fully agree about #36! =)

public/js/ClusterPlot.js
starcalibre: It's a constructor, and ``this`` is used to define properties of the object.

public/js/Histogram.js
starcalibre: Hahaha.. the package that lodash is forked from is called 'underscore'. Lodash has since surpassed underscore in pretty much every way, but underscore gets to keep the sensible name. :P

public/js/Histogram.js
starcalibre: This should probably be refactored like this..
```
var dataBinner = d3.layout.histogram()
    .bins(xScale.ticks(10))
var data = dataBinner(featureDist)
```

public/js/Histogram.js
starcalibre: 'g' is a tag used in SVG to group elements together. A transformation applied to the group will be applied to all elements in the group.

This grouping is applied to the axis because it's actually a whole bunch of SVG elements -- a line for the axis itself, and a whole bunch of lines and text for the ticks. By default, D3 will draw the x-axis along the top of the plot as SVG canvas have their origin in the top left. So the translate needs to be applied so the x-axis will draw along the bottom of the plot.

So basically what the code at line 73 is drawing is..

```
<svg>
    <g transformation=translate(0, xxx)> \\ xxx = whatever this.height equals  
         \\ bunch of SVG drawing the x-axis   \\ generated by .call(xAxis)
    </g>

No transformation is applied to the y-axis, but it's a good idea to put it in a group anyway in case any transformations need to be applied to it.

File: public/js/Lineplot.js (hunk 8 above)
Comment (starcalibre) [inline public/js/Lineplot.js]: Ack, missed some of these as you pointed out earlier.. I'll replace the other 'magic' height and widths you've pointed out so they're like this one. I think height, widths and target elements could be defined as arguments to the plot constructor, so something like this

```
function Plot(data, width, height, element) {
    this.fullWidth = width
    // etc
}
```

File: public/js/NeighbourPlot.js (hunk 11 above)
Comment (starcalibre) [inline public/js/NeighbourPlot.js]: lol, what did you want me to clarify here exactly?? :P

Comment (jni) [inline public/js/NeighbourPlot.js]: Well, 'updatePlot' is being assigned an empty function? What is the point of this?

Comment (jni) [inline public/js/NeighbourPlot.js]: And right under it it's being assigned a non-empty one! To reiterate: **WAT**

=P

File: public/js/Lineplot.js (hunk 8 above)
Comment (jni) [inline public/js/Lineplot.js]: Yes, I like that. But maybe make them later arguments so they can have default values? Also, is it possible to define these as percentage of width with some minimum width value? And would that be considered good practice? =P

File: public/js/Histogram.js (hunk 7 above)
Comment (jni) [inline public/js/Histogram.js]: This is really good! Could you add this to a comment either inline in the function, or at the top/bottom of the file, with a note inline to see the full reference? I just want to leave a good trail of breadcrumbs for our future selves.

File: public/js/Histogram.js (hunk 6 above)
Comment (jni) [inline public/js/Histogram.js]: Yeah that's not bad. =)

File: public/js/Histogram.js (hunk 5 above)
Comment (jni) [inline public/js/Histogram.js]: The creators of underscore also should be shot. The point is that an underscore is in no way a good package name. Might as well release "space", where it is imported as some unicode whitespace character that isn't ignored by parsers. Then it's all ` .function()` everywhere. FFS!

I get what they were going for, they were saying, "our library is going to be so fucking important that it won't need context, readable namespacing, or any of that stuff." Well, that's just arrogant and massively hurts readability.

What's `lodash-deep`? Does it somehow pollute the namespace even more? =P

File: public/js/Filter.js [hunk 13]
```diff
@@ -1,23 +1,26 @@
...
+
+function SampleFilter(sampleData) {
     $('.filter-items').children().remove();
 
-    this.columns = uniqueCol;
-    this.rows = uniqueRow;
-    this.plates = uniquePlate;
-    this.genes = uniqueGene;
+    this.data = sampleData;
+    this.uniqueCol = _.uniq(_.pluck(_.flatten(this.data), 'column')).sort();
```
jni: Maybe write a function for this?

```javascript
function unique_data(data, field) {
    return _.uniq(_.pluck(_.flatten(data), field)).sort()
}
```

public/js/Filter.js
```diff
@@ -1,23 +1,26 @@
-function SampleFilter(uniqueRow, uniqueCol, uniquePlate, uniqueGene) {
+var _ = require('lodash');
+
+function SampleFilter(sampleData) {
```
jni: I've finally figured out why I have so much trouble reading js code... Docstrings! I've been massively spoiled by Python docstrings. Is there an equivalent in js, and if so, can you sprinkle them all over this PR? In particular, for each function I would like:

- description of what the function does
- inputs (including type)
- outputs
- side effects (e.g. here, "adds xyz fields to the SampleFilter object")

starcalibre: When you say docstrings, do you mean something that adds a ``__doc__`` type property that can get parsed by a documentation package? If so.. no, this is not included in JS natively (there's no way Netscape would've thought of that back in 1995 :P), but there are modules that will do this -- JSDoc comes to mind. Seeing as the UI isn't an API, I don't there's any point in those types of docstrings?

But if you just mean detailed comments at the beginning of each the modules and their methods, then yes that would probably worth doing!

jni: Every function is an API, even if it is a private API! I've just glanced at JSDoc — is there any reason not to use comments structured as JSDoc expects, rather than your own syntax? It's kind of like NumPy style docstrings: even if you are not going to use the numpy docstring parser, they make the code that much more readable!

public/js/Histogram.js
starcalibre: ``lodash-deep`` adds a few methods to lodash that support operating on nested properties of objects. It adds a few different methods.. the 'circle slash circle' function ``%``, 'low-res snowflake' ``*`` and 'tic-tac-toe' ``#`` :P Kidding.. they're called from lodash, so they're just called from ``_``. See line 25:

``
var featureDist = _.deepPluck(self.sampleData,
        ['feature_vector_std', feature]);
``

This is reading the value of index at ``feature`` in the property ``feature_vector_std`` for every object in the ``sampleData`` array. It can access properties nested even further too. I was originally using my own function to do this, but I prefer this method. Say what you like about their naming conventions, the lodash devs are optimisation obsessed so I'll use their methods whenever appropriate!

public/js/Filter.js
starcalibre: I just had a look at the syntax myself. Looks good to me! I was more thinking there's no use *parsing* the doc strings as the UI methods won't be documented outside the source code.

public/js/NeighbourPlot.js
starcalibre: Ohhh, my bad -- forgot a todo. But they're actually two different functions -- ``updatePoint`` and ``updatePlot``.

``updatePoint`` is the function that controls the logic that changes the styling of the currently selected point, and updates the neighbour images gallery.
``updatePlot`` was going to be for the function that updates all of the plot points based on the filter parameters. I'll remove this because there's no benefit to keeping it there until it actually does something !

jni: Wow, amazing how the brain can be so stupid sometimes that I saw them as the same method name... =)

The distinction makes sense! =)

starcalibre:
@jni Changes:

* Amendments to the code based on your feedback and discussions above.
* Added a rough skeleton for the readme. I will flesh this out in a later commit.
* Added docstrings to the ClusterPlot object. Let me know what you think of this format and I'll write these for the rest of the plots too.

Oustanding: Finish writing the readme, finish the docstrings.

README.md
```diff
@@ -1 +1,70 @@
...
+--------------------
+
+```
+$ npm install
+$ bower install
+```
+
+Run dev version locally with
+```
+grunt
```
jni: `$ grunt`? For consistency with above. Additionally, you can get slightly nicer highlighting by adding "console" after the triple backticks.

README.md
```diff
@@ -1 +1,70 @@
...
+```
+
+MongoDB Schema
+--------------
+
+Screens saved in document with format:
+
+```
+var Screen = {
+    _id: String, // short screen name < 8 char
```
jni: What's this "< 8 char"??? Are we yearning for the good old days of MS-DOS???

README.md
```diff
@@ -1 +1,70 @@
...
+    screen_name: String, // full screen name
+    screen_desc: String,
+    number_samples : Number
+    features : Array // Array of features used in the screen
```
jni: are these the feature *names*?

README.md
```diff
@@ -1 +1,70 @@
...
+}
+```
+
+Each sample saved in document with format:
+
+```
+var Sample = {
+        _id: String, // name-plate-well
+        screen: String, // Screen _id
+        gene_name: String, // or gene_nane
```
jni: "or gene_nane" huh?

README.md
```diff
@@ -1 +1,70 @@
...
+
+```
+var Image = {
+        sample_id: String, // name-plate-well ENSURE A UNIQUE INDEX IS CREATED HERE!
+        image_full: String, // BinData of image
+        image_thumb: String // BinData of image (thumbnail)
+    }
+```
+
+Screen has a one to one relationship with Images.
```
jni: Do you mean Sample?

README.md
```diff
@@ -1 +1,70 @@
...
+
+Documents stored stored in the ``microscopium`` database, otherwise change specification in ``config/db.js``
```
jni: typo, "stored stored"

README.md
```diff
@@ -1 +1,70 @@
-Interface mockup
\ No newline at end of file
+# Microscopium User Interface
```
jni: Loving this README, thanks! =)

public/js/ClusterPlot.js
```diff
@@ -0,0 +1,170 @@
+var d3 = require('d3');
+require('d3-tip')(d3); // load d3-tip plugin
```
jni: maybe, `// load d3-tip plugin, adding it to the d3 namespace`?

public/js/Filter.js
```diff
@@ -43,15 +52,15 @@ function SampleFilter(uniqueRow, uniqueCol, uniquePlate, uniqueGene) {
...
 
     // append column checkboxes
-    var j = 0;
-    var cols = Math.ceil(uniqueCol.length/3);
-    for(var i = 0; i < uniqueCol.length; i++) {
-        var el = '#col-' + (j+1);
+    j = 0;
+    cols = Math.ceil(this.uniqueCol.length/3);
+    for(i = 0; i < this.uniqueCol.length; i++) {
+        el = '#col-' + (j+1);
```
jni: "el"? Like "El Chupanigre?" =P Maybe expand this variable name?

public/js/Histogram.js
```diff
@@ -0,0 +1,100 @@
+var d3 = require('d3');
+var _ = require('lodash');
+_.mixin(require('lodash-deep'));
+
+function Histogram(sampleData, element, featureNames) {
```
jni: Ech. Is there really no "class" declaration keyword in JS??? =(

public/js/NeighbourPlot.js
```diff
@@ -0,0 +1,199 @@
+var d3 = require('d3');
+require('d3-tip')(d3);
+
+/**
+ * Draws PCA scatterplot of the data Clicking on scatterplot points will
```
jni: I prefer the imperative, as in Python: "Draw PCA scatterplot of the data"

Couple of newlines missing?

Finally, let's keep it general, because I want to try tSNE soon: "Draw scatterplot of the dimension-reduced data"

public/js/Histogram.js
starcalibre: No, JavaScript has no class :P It's a prototypical inheritance language. So these plots aren't defined as their own class, they are extensions of a base ``Object`` class.

public/js/NeighbourPlot.js
jni: How about adding the function name here? Man Python really got it right with their docstrings, they should definitely come *after* the function name! I started reading this commend going huh? wtf is he talking about? Then I scrolled down and realised it was a JSDoc! How about:

```
/**
 * NeighbourPlot: class to draw scatterplot of dimension-reduced data
 *
 * Clicking on scatterplot points will trigger updates of the neighbour image
 * [...]
```

What do you think?

public/js/NeighbourPlot.js
```diff
@@ -0,0 +1,199 @@
...
+
+    this.margin = {top: 10, right: 40, bottom: 30, left: 40};
+    this.width = this.fullWidth - this.margin.left - this.margin.right;
+    this.height = this.fullHeight - this.margin.top - this.margin.bottom;
+
+    this.drawScatterplot();
+}
+
+/**
+ * This function draws the PCA plot.
```
jni: function -> method? This would keep it consistent with common OO terminology and help my brain to not hurt. =P

public/js/NeighbourPlot.js
```diff
@@ -0,0 +1,199 @@
...
+    var firstPoint = d3.select('.scatterpt');
+    self.updatePoint(firstPoint, firstPoint.data()[0],  0);
+};
+
+/**
+ * This function is attached to all PCA scatterplot points, and updates
+ * the styling of the points and triggers updates in the associated samples
+ * lineplot and neighbour image gallery.
+ *
+ * @this {NeighbourPlot}
```
jni: Add parameters

public/js/NeighbourPlot.js
```diff
@@ -0,0 +1,199 @@
...
+ *
+ * @constructor
+ * @param {array} sampleData - The sample data for the screen. Each element
+ * in the array is an instance of a Sample document.
+ * @param {string} element - The ID of the target div for this plot.
+ * @param {LinePlot} lineplot - LinePlot that will update when scatterplot
+ * points are clicked.
+ * @param {NeighbourImages} neighbourImages - NeighbourImages object that will
+ * update when scatterplot points are clicked.
+ */
```
jni: Overall, this is a great addition! =)

public/js/Histogram.js
jni: What does "prototypical inheritance" mean, other than being a cool name for a videogame? =P But, to reiterate, "ech". Point is, there's no real way to tell whether something is an "object" or a plain function until you start seeing "this" everywhere, right?

starcalibre: Prototypical inheritance just means that objects inherit from other objects. This contrasts to classical inheritance where classes inherit from other classes, and these classes in turn define objects. Prototypical inheritance has some advantages -- it's simple yet powerful. It's possible to easily extend objects without a whole bunch of extra infrastructure code. This is useful in a dynamic environment like JavaScript. Everything just inherits from an ancestor object.

Say I want to extend the Array object to add this awesome new method:

```
> Array.prototype.sayHello = function() {
>    console.log('I\'m just an array, please don\'t slice and splice me!');
> }
> [1, 2, 3, 4].sayHello()
I'm just an array, please don't slice and splice me!
```

In the classical inheritance model, I'd need to define a new class that extends the Array class to do that, THEN create a new instance of that new extended class.

Technically, JavaScript actually implements a *pseudo-classical* inheritance model. That's why there's the constructor function. It's not really 'constructing' anything in the traditional sense where a class definition is used to build a new object. It's just creating an object, and putting all those properties you've defined into it. Then when the function is called using the ``new`` keyword, a new instance of that object is created, ie the new object inherits from an ancestor object. 

These "constructors" are differentiated from normal functions by capitalising the first letter of their name. Ditto how classes are named in other OOP languages. This isn't enforced, but it's pretty much universally accepted as the way to do it and will be enforced by default in all JS linters.

ES6, the new version of JavaScript is going to introduce a ``class`` keyword. It's just syntactic sugar and doesn't change the fundamental nature of JS but I'm not big on it. I think it'll just confuse matters more, especially if you're a programmer coming from a classical inheritance background.

public/js/NeighbourPlot.js
starcalibre: I've been reading up on tSNE. Looks great. I think a nice addition would be the option to give the user a few options about the dimension is reduced to draw the scatterplot. A dropdown box with PCA, MDS, tSNE or any others are are applied.

Regarding the docstrings, how strict we adhere to the JSDoc syntax all depends on whether or not we want to parse them. I don't think we will, so I'm happy to add a little more detail as you've described above.

public/js/Histogram.js
jni: This is gold! Thank you! =)

jni: btw, do you have a blog? You should: it's really great for exposure — I'm pretty sure that's how I got my book deal! Regardless, you should write a blog post that copies what you just wrote almost verbatim. =)

It is also really good practice to write often, and then you can pilfer it when you are writing your thesis!

public/js/NeighbourPlot.js
jni: I'd bury the different dimensionality reduction options deep in an "advanced" menu somewhere... I think we can do pretty well ourselves figuring out how we want to show the data, and it's good not to overwhelm newbies with too many options!

As to the docstring, I specifically tried to design something that would not be too egregious in violating JSDoc syntax. I think the above would render fine *even if* we eventually compile the JSDoc to HTML; it would just be a bit repetitive.

public/js/Histogram.js
jni: Also, don't sanitise the story and make it a sterile technical manual about classes and objects in JS. Start with something like, "My MSc advisor, @jni, is totally clueless when it comes to JS, so he asked me this noob question the other day..." Ok maybe not exactly that but you get my drift. I hate that scientific writing tries to erase the humans from the equation. It's supposed to make it more "objective", but clearly it is worthless since something like 60-90% of scientific studies don't reproduce... So imho you might as well bring back the human element.

`</rant>`

starcalibre:
@jni -- that should cover everything discussed above!

README.md
jni: Still need to add these:

    ```console
    $ grunt
    ```
And add the `console` language keyword to the npm/bower lines.

README.md
jni: Still wondering about this screen name limitation?

README.md
jni: Array of String? If this is indeed the feature names, please specify in the comment string.

README.md
jni: Still haven't explained this?

README.md
jni: This still needs fixing

README.md
jni: typo still not fixed...

public/js/ClusterPlot.js
```diff
@@ -0,0 +1,212 @@
...
+ * ClusterPlot: Object to draw scatterplot of dimension reduced data.
+ *
+ * This scatterplot is controlled by a slider that allows the user to
+ * select the number of k partitions they would like the data clustered into.
+ * The plot points are coloured according to which cluster they belong to for
+ * the currently selected k.
+ *
+ * @constructor
+ * @param {array} sampleData - The sample data for the screen. Each element
+ * in the array is an instance of a Sample document.
```
jni: Can you indent this four spaces for better visual parsing of all the parameters? e.g.:

```
 * @param {array} sampleData - The sample data for the screen. Each element
 *     in the array is an instance of a Sample document.
 * @param {string} element - The ID of the target div for this plot.
```

public/js/ClusterPlot.js
```diff
@@ -0,0 +1,212 @@
...
+ * @param {string} element - The ID of the target div for this plot.
+ */
```
jni: Other than that minor formatting quibble, great docstring! I really like how it reads and the title line really prepares me for what's coming.

public/js/ClusterPlot.js
```diff
@@ -0,0 +1,212 @@
...
+};
+
+/**
+ * redrawClusters: Update the scatterplot points.
+ *
+ * When a new value of k is chosen, update the colouring of the
+ * scatterplot points and their tool-tips to show which cluster they
+ * belong to.
+ *
+ * @this {redrawClusters}
```
jni: I'm confused: doesn't `this` stand for a `ClusterPlot` instance here?

public/js/ClusterPlot.js
```diff
@@ -0,0 +1,212 @@
...
+        })
+        .on('mouseover', tip.show)
+        .on('mouseout', tip.hide)
+        .call(tip);
+};
+
+/**
+ * destroy: Remove all child SVG elements of the plot objects containing div.
+ *
+ * @this {Lineplot}
```
jni: Is `Lineplot` a d3 class? If so, is this better expressed as `d3.Lineplot`?

public/js/Filter.js
```diff
@@ -1,24 +1,43 @@
-function SampleFilter(uniqueRow, uniqueCol, uniquePlate, uniqueGene) {
+var _ = require('lodash');
+
+/**
+ * SampleFilter: Setup sample filter.
+ *
+ * The constructor function adds all filter filters based on the current
```
jni: I'm glad it only adds filter filters and not other kinds of filters. =P

public/js/Filter.js
```diff
@@ -1,24 +1,43 @@
...
+ * sample data and attaches event listeners that drive behaviour filter.
+ *
+ * @constructor
+ * @param {array} sampleData - The sample data for the screen. Each element
+ * in the array is an instance of a Sample document.
```
jni: As before, please indent subsequent lines of a docstring parameter description.

public/js/Filter.js
```diff
@@ -120,21 +139,14 @@ function SampleFilter(uniqueRow, uniqueCol, uniquePlate, uniqueGene) {
     $('#filter-button').removeClass('hidden');
 }
```
jni: This function is enormous! Can you manage to break it down a bit, farm out a few of those loops to some helper functions? (and I think this closing curly brace is misaligned LOL)

public/js/ClusterPlot.js
jni: Nevermind, I see now Lineplot is your own. But is it the right `this`, here? Or just the result of some over-eager copy-pasting? ;)

public/js/NeighbourPlot.js
```diff
@@ -0,0 +1,208 @@
...
+        .style('text-anchor', 'middle')
+        .text('PC2');
+
+    // default select first point
+    var firstPoint = d3.select('.scatterpt');
+    self.updatePoint(firstPoint, firstPoint.data()[0],  0);
+};
+
+/**
+ * updatePoint: Update catterplot point on click.
```
jni: Is a catterplot a plot full of kittehs? =D

jni:
Epic PR! There's still a few minor comments to address (and one major — refactoring that huge constructor to use a few more functions), but overall it's looking great!

starcalibre:
@jni new commits ! :)

jni:
I don't understand why the filter constructor refactoring didn't result in more deleted lines...?

jni:
Oh, it's in the previous commit! LOL Ok gotcha. Well, still another round, but it should be quick! Much improved otherwise, thanks!

jni:
And this still appears to be in there:

"Screen has a one to one relationship with Images"

`s/Screen/Samples/` ?

starcalibre:
@jni ping!

This was a little more work than I thought it would be, but it's a worthwhile change I think. This is *nearly* done.. there's one thing outstanding, rather than use the ``numberCols`` parameter to decide how the filter menus will display.. I think this should be determined by the number of unique values the menu needs to display.

jni:
Fan-friggin-tastic. That looks way better. Thanks! I would say your final task can probably wait for another PR? Would you like me to merge? I'd like to get the actual filtering going! =)

starcalibre:
Go for it. I'll work on adding the filter functionality asap!

jni:
hashtag boom. =D

PR #41, opened 2015-03-30
Add installation instructions for the grunt CLI

jni:
This would have prevented some minor confusion for me earlier today. =) But let me know if it's not the recommended way to install this. (Trying to install it without the `-g` flag stalled for me.)


starcalibre:
Thanks for this!

PR #43, opened 2015-04-17
Improve performance of neighbour plot

starcalibre:
* Use CSS instead of SVG attributes to style points.
* The updatePlot method now moves unfiltered points to the top of the SVG pile in one function call.

This is running insanely fast in Chrome, and much much better in Firefox.


jni:
That is *insane*. I can confirm the massive speedup in Chrome and the noticeable speedup in Safari. Awesome!

PR #44, opened 2015-04-20
Update express and middlewares

starcalibre:
* Use the latest version of the express framework
* Use latest version of morgan and bodyparser
* Add the compress middleware -- uses gzip when sending responses from the server. Given the server will be sending lots of JSON (a plain-text format), it makes lots of sense to use this. At the moment it's just using the default settings which applies a mid-level of compression to all responses, but these can be customized so more compression is used for larger responses.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,15 +2,16 @@
     "name": "microscopium-ui",
     "main": "server.js",
     "dependencies": {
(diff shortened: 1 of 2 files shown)


jni:
Awesome, thanks! =)

PR #45, opened 2015-04-27
[WIP] Refactor interface to show all plots in one tab

starcalibre:
I've marked this WIP because there's still a couple of bits and pieces to be done, such as updating the docstrings to relfect the changes I've made. However the functionality is there if we want to try putting the branch online.

Other updates in this PR include:
* If only one screen is found in the Mongo collection, this is loaded automatically.
* Clicking on the images triggers an update of the active selected point, so its position in the PCA plot and feature distribution in the linegraph are displayed.
* When a point is clicked, it's brought to the top of the SVG stack.
* A few quick optimizations, such as caching jQuery selectors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ var Image = {
         _id: ObjectID, // Mongo generated unique ID for this image
         sample_id: String, // The _id of the sample corresponding to this image.
         image_full: String, // The fullsize image.
-        image_full: String, // A large preview image, 512px wide.
+        image_large: String, // A large preview image, 512px wide.
         image_thumb: String // A smalller, thumbnail sized preview image, 150px wide.
     }
 ```
--- a/README.md
+++ b/README.md
@@ -99,7 +99,8 @@ Each image saved in document with format:
 var Image = {
         _id: ObjectID, // Mongo generated unique ID for this image
         sample_id: String, // The _id of the sample corresponding to this image.
-        image_full: String, // BinData of image
-        image_thumb: String // BinData of image (thumbnail)
+        image_full: String, // The fullsize image.
+        image_full: String, // A large preview image, 512px wide.
+        image_thumb: String // A smalller, thumbnail sized preview image, 150px wide.
     }
 ```
--- a/app/routes.js
+++ b/app/routes.js
@@ -62,6 +62,21 @@ module.exports = function(app) {
             .exec(resHandler(res))
     });
 
+    // get all neighbours
+    app.get('/api/sample/neighbours/:id', function(req, res) {
+        Sample
+            .find({ '_id': req.params.id })
+            .select('neighbours')
+            .exec(function(err, data) {
+                Sample.find({
+                    '_id': { $in: data[0].neighbours }
+                })
+                .exec(function(err, data) {
+                    res.send(data);
+                });
+            })
+    });
+
     // get value from sample document
     app.get('/api/sample/:key/:id', function(req, res) {
       Sample
@@ -78,14 +93,19 @@ module.exports = function(app) {
         .exec(resHandler(res));
     });
 
-    // get image thumbnails
-    app.get('/api/images/', function(req, res) {
-      Image
-        .find({
-          'sample_id': { $in: req.query.sample_ids }
-        })
-        .select('image_thumb sample_id')
-        .exec(resHandler(res));
+    // get thumbnails for all neighbour images
+    app.get('/api/images/:id', function(req, res) {
+        Sample
+            .find({ '_id': req.params.id })
+            .select('neighbours')
+            .exec(function(err, data) {
+                Image
+                    .find({
+                        'sample_id': { $in: data[0].neighbours }
+                    })
+                    .select('sample_id image_thumb')
+                    .exec(resHandler(res));
+            })
     });
 
     // default route
--- a/app/routes.js
+++ b/app/routes.js
@@ -51,7 +51,7 @@ module.exports = function(app) {
         Sample
         .find({ 'screen': req.params.screen })
         .select('_id row column plate gene_name feature_vector_std ' +
-                'pca cluster_member neighbours')
+                'pca_vector neighbours')
         .exec(resHandler(res));
     });
 
@@ -85,11 +85,11 @@ module.exports = function(app) {
         .exec(resHandler(res))
     });
 
-    //get fullsize image
+    //get large image
     app.get('/api/image/:id', function(req, res) {
         Image
             .find({ 'sample_id': req.params.id })
-        .select('sample_id image_full')
+        .select('sample_id image_large')
         .exec(resHandler(res));
     });
 
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -21,6 +21,10 @@
     margin-top: 50px;
 }
 
+.sb-active {
+    overflow: hidden;
+}
+
 .panel {
     background-color: #222222;
 }
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -81,15 +81,15 @@
 .thumb {
     margin: 1px 1px -3px 1px;
     padding: 0 0 0 0;
-    width: 150px;
-    height: 150px;
+    width: 100px;
+    height: 100px;
 }
 
 .selected-image-display {
     margin: 1px 1px -3px 1px;
     padding: 0 0 0 0;
-    width: 300px;
-    height: 300px;
+    width: 250px;
+    height: 250px;
 }
 
 /* Centered columns for grid display in Nebula */
@@ -107,7 +107,7 @@
 
 /* Plot all D3 SVGs to center of their divs */
 .plotbox {
-    text-align: center;
+    text-align: left;
 }
 
 /* GENERAL PLOT CSS */
@@ -147,7 +147,7 @@
 /* NEIGHBOUR PLOT CSS */
 
 .gallery {
-    height: 500px;
+    height: 675px;
     overflow: scroll;
 }
 
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -1,5 +1,10 @@
 /* MAIN PAGE CSS */
 
+#image-column {
+    overflow: scroll;
+    height: 0;
+}
+
 .load-overlay {
     pointer-events: none;
 }
@@ -85,15 +90,11 @@
 .thumb {
     margin: 1px 1px -3px 1px;
     padding: 0 0 0 0;
-    width: 100px;
-    height: 100px;
 }
 
 .selected-image-display {
     margin: 1px 1px -3px 1px;
     padding: 0 0 0 0;
-    width: 250px;
-    height: 250px;
 }
 
 /* Centered columns for grid display in Nebula */
@@ -149,11 +150,6 @@
 }
 
 /* NEIGHBOUR PLOT CSS */
-
-.gallery {
-    height: 675px;
-    overflow: scroll;
-}
 
 .pcaplot {
     height: 500px;
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -106,11 +106,6 @@
     text-align: center;
 }
 
-/* Plot all D3 SVGs to center of their divs */
-.plotbox {
-    text-align: left;
-}
-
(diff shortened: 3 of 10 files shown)


README.md
```diff
@@ -99,7 +99,8 @@ Each image saved in document with format:
 var Image = {
         _id: ObjectID, // Mongo generated unique ID for this image
         sample_id: String, // The _id of the sample corresponding to this image.
-        image_full: String, // BinData of image
-        image_thumb: String // BinData of image (thumbnail)
+        image_full: String, // The fullsize image.
+        image_full: String, // A large preview image, 512px wide.
```
jni: Rather repetitive, don't you think? =P `image_large`?

app/routes.js
```diff
@@ -62,6 +62,21 @@ module.exports = function(app) {
             .exec(resHandler(res))
     });
 
+    // get all neighbours
+    app.get('/api/sample/neighbours/:id', function(req, res) {
```
jni: This is basically exactly what I'm talking about with anonymous functions. This should read:

```javascript
app.get('api/sample/neighbours/:id', get_neighbours(req.params.id, res))
```

Later, you can do something similar with getting the thumbnails.

jni: and the images. Basically, there's no reason for these gets to be longer than one line.

public/css/style.css
```diff
@@ -78,18 +87,10 @@
...
-    width: 150px;
-    height: 150px;
-}
-
-.selected-image-display {
-    margin: 1px 1px -3px 1px;
-    padding: 0 0 0 0;
-    width: 300px;
-    height: 300px;
+    height: 100%;
```
jni: Hmm, do we really not want to specify a maximum width here?

jni: Or will this be the width of the enclosing element by default?

public/css/style.css
```diff
@@ -107,7 +108,7 @@
 
 /* Plot all D3 SVGs to center of their divs */
 .plotbox {
-    text-align: center;
+    text-align: left;
```
jni: LOL, look at the comment above, and then at [this image](https://twitter.com/nzkoz/status/538892801941848064). =D

public/index.html
```diff
@@ -57,207 +45,172 @@
...
+                        <div class="col-lg-12 col-centered sample-image">
                             <a class="thumbnail" href="#">
-                                <img id="nebula-0" class="img-responsive" src="" alt="">
+                                <img id="nebula-0" src="" alt="">
                             </a>
                         </div>
                     </div>
-                    <div class="row-centered">
-                        <div class="col-lg-12 col-centered thumb">
+                    <div class="row row-centered">
```
jni: This is actually hideous... =( Is there no way to auto-generate this code? (Sorry, I know we had this discussion but I've forgotten the outcome. Perhaps I blocked it out of my brain. =P)

public/js/Histogram.js
```diff
@@ -12,21 +16,28 @@ _.mixin(require('lodash-deep'));
  * @param {array} featureNames - An array of the features used in this screen.
  */
 function Histogram(sampleData, element, featureNames) {
+    var self = this;
```
jni: Why is this line needed? Why not just use this everywhere? (which you mostly do)

public/js/Histogram.js
```diff
@@ -12,21 +16,28 @@ _.mixin(require('lodash-deep'));
...
+
     this.sampleData = sampleData;
     this.featureNames = featureNames;
     this.feature = 0;
-
     this.element = element;
-    this.fullWidth = 500;
-    this.fullHeight = 400;
+
+    this.fullWidth = $(this.element).width();
```
jni: Can you add a comment explaining what $ does? Lame for expert js-ers but I know I'll be grateful one day. =)

public/js/Lineplot.js
```diff
@@ -8,29 +8,43 @@ var _ = require('lodash');
  * lineplot triggers an update in a Histogram object showing the distribution
  * of the corresponding feature.
  *
+ * The width and height of the plot is calculated based on the size of the
+ * plot's containing DIV. The width is taken from the DIV, and the height
+ * calculated from that width such that the plot will have a 16:9 aspect ratio.
+ *
```
jni: Very nice comment! =)

public/js/Lineplot.js
```diff
@@ -199,14 +212,12 @@ Lineplot.prototype.keypressUpdate = function(keyCode) {
 
     if(self.activeFeature) {
         if(keyCode === 39 && self.activeFeature < this.featureLength) {
-            this.activeFeature++;
-            self.histogram.drawHistogram(self.activeFeature-1);
-            self.updateActiveLine(self.activeFeature);
+            self.activeFeature++;
+            $('body').trigger('linePlotUpdate', self.activeFeature);
         }
         else if(keyCode === 37 && self.activeFeature > 1) {
```
jni: Ooh, can you create some constants for these, say LEFTARROW and RIGHTARROW? It's both more readable and less brittle.

public/js/NeighbourImages.js
```diff
@@ -54,18 +73,54 @@ NeighbourImages.prototype.getImages = function(query_id) {
...
 NeighbourImages.prototype.setImages = function() {
     var self = this;
+    var $nebula0 = $('#nebula-0');
+
+    // make all img frames empty
+    $nebula0
+        .attr('src', '//:0')
+        .attr('title', 'not found');
+
+    $('.thumbnail > img')
```
jni: This syntax is really throwing me off. Can you point me to a writeup on it? (Either on our repo or in the wider web, linked from our repo.)

public/js/NeighbourPlot.js
```diff
@@ -27,23 +27,25 @@ d3.selection.prototype.moveToFront = function() {
...
+function NeighbourPlot(sampleData, element) {
+    var self = this;
 
+    this.sampleData = sampleData;
     this.element = element;
-    this.fullWidth = 700;
-    this.fullHeight = 500;
+
+    this.fullWidth = $(this.element).width();
+    this.fullHeight = Math.round(this.fullWidth * (9/16));
```
jni: Good stuff! =)

jni: Reminds me of something I saw a while back: "there's two kinds of people in this world: those who understand resolutions and aspect ratios; and morons." =D I still marvel at all the *brand new planes* with 4:3 movies shown stretched on a 16:9 screen. Who the fuck is responsible for these things???

Ok that was slightly off topic. =P

public/js/Histogram.js
starcalibre: Creating a reference to the ``this`` object is useful for when I went to reference to that particular ``this`` when I'm not in it's scope.

Here's an example from ``lineplot.js``
````
var line = d3.svg.line()
    .x(function(d) { return self.xScale(d[0]); })
    .y(function(d) { return self.yScale(d[1]); });
```

When I call the on method, if I used the ``this`` keyword instead of the ``self` variable I've created, ``this`` will point to the anonymous function rather than the plot object so this won't work.

There is a ``bind`` method, so the above code segment could be written like so:
````
var line = d3.svg.line()
    .x(function(d) { return self.xScale(d[0]); }.bind(this))
    .y(function(d) { return self.yScale(d[1]); }.bind(this));
```
But I personally find it hideous. And there's also situations where you want to call the ``this`` of both the closure you're in, *and* the ``this`` of the other enclosures. See this below example:

````
self.svg = d3.select(this.element).append('svg')
    .attr('width', self.fullWidth)
    .attr('height', self.fullHeight)
    .append('g')
    .attr('transform', 'translate(' + self.margin.left + ',' + self.margin.top + ')')
    .on('click', function() {
        self.onClickUpdate(d3.mouse(this));
    });
````
When handling the click event, I want to call the method ``onClickUpdate`` the belongs to the outer scope, *and* call ``d3.mouse(this)``. If used bind, the wrong reference would get passed to get ``d3.mouse``. 

Open to any suggestions! Like everything in JS, there's no convention. :P

File: public/js/Histogram.js (hunk 7 above)
Comment (starcalibre) [inline public/js/Histogram.js]: ``$`` is just a reference to the jQuery object. It's exactly the same as calling ``jQuery(this.element)`` but the shorthand is always used because jQuery is so ubiquitous it's basically part of Javascript.

File: public/js/Histogram.js (hunk 6 above)
Comment (jni) [inline public/js/Histogram.js]: Got it! Suggestion: copy this comment verbatim and add it somewhere in the code! =) Totally get it now.

It's not making me any more enamoured of anonymous functions, though... =P

File: public/js/Histogram.js (hunk 7 above)
Comment (jni) [inline public/js/Histogram.js]: Good God! I finally get it! This is even more egregious than lodash! LOL

What about `$foo`? Does that still "call" jQuery? (ie parens are optional?)

File: public/js/script.js [hunk 12]
```diff
@@ -32,6 +35,7 @@ function updateSelector(screen_data) {
     }
 }
 
+// FIXME does not corectly switch between screens
```
jni: [ROTFL](http://www.bonkersworld.net/images/2012.08.15_todo.png)

(plus you made a spelling mistake in your fixme comment! LOL)

jni:
Okely! Done! Essentially I've asked you to remove a couple of anon functions but otherwise pretty good! Not really that much code changed for a hugely different functionality! =)

public/js/Histogram.js
starcalibre: It's a jQuery object that's been stored into a variable. If you're going to select an element more than once, it's good practice to cache the selector into a variable. That way, jQuery doesn't need to traverse the DOM to find the element again. It's stored in a variable so you can access it again immediately.

The $ is there just to show it's a jQuery object -- that's a convention set out in the AirBnb style guide! https://github.com/airbnb/javascript#jquery

public/js/script.js
starcalibre: haha, that's fixed now! So I'll add a TODO comment to the remove FIXME coment ;) :P

public/js/NeighbourPlot.js
starcalibre: My first job was at a Blockbuster Video back when those were a thing.. part of me died every time somebody complained 'this tap has those annoying black bars that cut off part of the screen!' !

public/js/Histogram.js
jni: I love than there's an AirBnB style guide for JS... BnB's are associated in my brain with nice little old ladies — not your typical JS hacker...! LOL

Thanks for the info!

public/js/script.js
jni: :+1: !

public/js/NeighbourPlot.js
jni: I also remember reading an article back in the early days of DVDs about a guy whose aunt would always get him the "full screen version" because she knew he cared about that sort of thing — of course the marketing morons at video distributors had called 4:3 "full screen" because they are morons. Morons! LOL

PS: that is absolutely awesome that you worked at a Blockbuster. =D

jni:
Incidentally, for all the bits where I say "great comment, can you document it somewhere", but there's no obvious place in-line in the code for it, how about adding a "notes" directory where you include all this in text files? I just saw that we have that for microscopium and I think it's a good idea, to help new coders get up to speed on our (your) conventions...!

public/css/style.css
starcalibre: Hah, text align defaults to left anyway so this class can go!

public/index.html
starcalibre: Two options -- render it server-side using handlebars or jsdom, or render it client-side using jQuery. I agree it can be reduced but I'd rather make this something to do last.. if I need to change the way the gallery is displayed for some reason, it's much easier having the whole HTML layout right there.

app/routes.js
starcalibre: The next PR (faster loading) is going to include quite a few changes to the DB API -- simplying the routes and making them more general. I'll address these comments in that PR.

README.md
jni: @starcalibre this still needs fixing! (`image_full` -> `image_large`)

jni:
Exciting!

PR #46, opened 2015-04-28
[WIP] Update server/configs before putting demo online

starcalibre:
I don't want to merge this just yet, but I'll use this branch to track any updates made to the backend before merging to master and putting the demo online.

* The port is read from the PORT env variable, otherwise it defaults 8080 if this isn't set.
* The Mongo URL, username and password are read from env variables -- if these aren't set, then it'll try and connect to the unsecured ``mongodb://localhost/microscopium`` instance.

### DIFF
--- a/config/db.js
+++ b/config/db.js
@@ -1,4 +1,6 @@
 module.exports = {
     // the database url
-    url : 'mongodb://localhost/microscopium'
-}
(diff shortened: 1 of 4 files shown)


jni:
This looks great, let me know when you want me to hit the merge button!

PR #51, opened 2015-05-12
Faster initial load

starcalibre:
Here are the changes from today.
* Update the UI to use the feature document to get the distribution of a feature. This prevents the need for the UI to download the whole dataset when it's initially loaded.
* Update the readme to describe the feature document, and added some info about which indices to add to the DB.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,11 @@ MongoDB Schema
 Documents should by default be stored in the the database ``microscopium``,
 however this can be changed by updating the ``config/db.js`` file.
 
+### Document Schemas ###
+
+While MongoDB does not enforce a schema for data, certain fields must be present in
+each document to ensure proper operation of the UI.
+
 Screen descriptions are saved in the ``screens`` collection with the
 following format:
 
@@ -103,4 +108,30 @@ var Image = {
         image_large: String, // A large preview image, 512px wide.
         image_thumb: String // A smalller, thumbnail sized preview image, 150px wide.
     }
+```
+
+For each feature belonging to each screen, a document is saved detailing the feature
+across the whole dataset. Think of this is the column axis if you were to look at the
+dataset as a dataframe or table.
+
+```
+var Feature = {
+        _id: ObjectID, // Mongo generated unique ID for this feature.
+        screen: String,
+        feature_name: String, // the actual name of the feature
+        feature_dist: Array, // A vector of unstandardised values for this feature.
+        feature_dist_std: Array, // A vector of standardised values for this feature.
+}
+```
+### Indices ###
+
+By default, MongoDB only creates indices on the primary key for each document.
+The UI makes queries to a few non-primary keys, so indices need to be created on these.
+From your Mongo terminal, load up the ``microscopium`` database and enter these commands
+to ensure you've got the right indices.
+
+```
+db.samples.createIndex({'screen':1});
+db.images.createIndex({'sample_id':1});
+db.features.createIndex({'screen':1, 'feature_name':1});
 ```
--- a/app/routes.js
+++ b/app/routes.js
@@ -1,6 +1,10 @@
+var Feature = require('./models/feature');
 var Screen = require('./models/screen');
 var Sample = require('./models/sample');
 var Image = require('./models/image');
+
+var _ = require('lodash');
+_.mixin(require('lodash-deep'));
 
 var resHandler = function(res, sort) {
     if(sort === true) {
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -52,6 +52,11 @@ MongoDB Schema
 Documents should by default be stored in the the database ``microscopium``,
 however this can be changed by updating the ``config/db.js`` file.
 
+### Document Schemas ###
+
+While MongoDB does not enforce a schema for data, certain fields must be present in
+each document to ensure proper operation of the UI.
+
```
jni: Ooh I like this paragraph! =)

app/routes.js
```diff
@@ -24,21 +28,14 @@ module.exports = function(app) {
...
     });
 
-    // get all screen ids
-    app.get('/api/screen_ids', function(req, res) {
+    // get fields of all screens
+    app.get('/api/screens/:field/', function(req, res) {
         Screen
             .find({})
-            .select('_id')
+            .select(req.params.field)
```
jni: What's a field?

jni: definitely needs clarification — can only think of microscope tiles, which I don't think is what you're talking about here!

app/routes.js
```diff
@@ -46,43 +43,50 @@ module.exports = function(app) {
...
-    app.get('/api/sample/:id', function(req, res) {
-          Sample
-            .find({ '_id': req.params.id })
-            .exec(resHandler(res))
+    // get value from sample document
+    app.get('/api/sample/:id/:field', function(req, res) {
+      Sample
+        .find({ '_id': req.params.id })
+        .select(req.params.field)
+        .exec(resHandler(res))
```
jni: Hey look, these are exactly the same lines as 49-51! -> Refactor. =) I see that "sample" and "screen" are different, but you can parameterise on that. Approximate example:

```javascript
get_field_callback = function(collection) {
    return function(req, res) {
        collection.find({}).select().exec() }

app.get('/api/sample/:id/:field', get_field_callback(Sample))`
```

app/routes.js
```diff
@@ -46,43 +43,50 @@ module.exports = function(app) {
...
-                .exec(function(err, data) {
+                if(req.query.pluck) {
+                    if(err) {
+                        res.send(err);
+                    }
+                    var pluck = _.deepPluck(data,
+                        [req.params.field, req.query.pluck]);
+                    res.send(pluck);
+                }
+                else {
```
jni: Baaaah you're killing me! Anon functions three levels deep!!!! =(

public/js/Lineplot.js
```diff
@@ -214,13 +201,14 @@ Lineplot.prototype.keypressUpdate = function(keyCode) {
     var self = this;
 
     if(self.activeFeature) {
-        if(keyCode === RIGHTARROW && self.activeFeature < this.featureLength) {
+        if(keyCode === RIGHTARROW &&
+            self.activeFeature < this.featureVector.length) {
```
jni: Is this 0- or 1-indexed? If 0, won't this fail when we get to the rightmost feature and try to increment it?

jni:
Done with comments! I'm going to be a pain and ask for fewer anon functions where it is at all possible!

jni:
Oh one more comment: at some point (after I've cleaned up feature computation for problematic images), instead of loading the full feature distribution, we can reduce it even further by loading some distribution descriptors, e.g. all the percentiles, or just a histogram with 100 bins. This can get added here or just as an issue.

public/js/Lineplot.js
starcalibre: The array of feature vectors is 0-indexed, but in the plot they're 1-indexed.. i.e. feature 0 maps to the x=1 co-ordinate of the line plot, feature 1 maps to x=2 etc. 

I had to think about this for more than 30 seconds, so I better add a comment explaining the indexing !

app/routes.js
starcalibre: It's a field in the context of database.. i.e. a column name. This is what they're referred to in the MongoDB documentation. I think they avoid the use of the word "key" as that usually refers to unique identifiers such as ``_id``.

app/routes.js
starcalibre: I've created issue #52 for this !

jni: @starcalibre are you saying you're punting on not displeasing me with the huge number of anons? =)

Incidentally, #52 becomes much, much easier if you create functions (only need to modify the function) vs using anons (need to modify each and every anon).

jni:
@starcalibre let me know when/if you think this is ready for another round of revision. If you push me I'll merge with the anons in there, but I consider it technical debt and I think we have quite a bit of it already! (Search "US government shutdown 2014" for examples of how this can escalate... =P)

starcalibre:
@jni Hey I'm in the middle of rewriting the API, routes and AJAX request accordingly. This. is. looking. AWESOME. It'll be a substantial bunch of commits though.. would it be easier for you look over if it was in a separate PR?

jni:
Nah, just pile it on to this one — it'll make me happier. =)

jni:
btw, sounds very exciting! =D

starcalibre:
@jni Okay, this is good to go! I think the comments and code make it fairly obvious where I'm going with it but let me know if anything isn't clear :)

The DB API still has lots more functionality to add, such as support for more sophisticated queries but this is a good foundation for now.

jni:
@starcalibre I've made some comments on the individual commits (easier to review). This is *way* better! Thanks! (And great new docs!) Only a couple of minor things and it's going in. Brilliant! =)

starcalibre:
@jni This should be good to go !

jni:
Music to my eyeballs! ... Or something.

Looks wonderful. I'm very happy with the refactor. In it goes!

PR #61, opened 2015-05-19
Add backwards and forwards buttons

starcalibre:
This PR adds buttons to the top menu bar that let a user go backwards and forwards between samples they've chosen.
backwards-forwards

### DIFF
--- a/public/js/History.js
+++ b/public/js/History.js
@@ -7,7 +7,8 @@
  */
 function History() {
     this.sampleHistory = [];
-    this.iterator = 0;
+    this.current = -1;
+    this.lastSample = null;
     this.length = 0;
 }
 
@@ -20,17 +21,21 @@ function History() {
  * @param {string} sampleId - The sampleID to add to the history.
  */
 History.prototype.add = function(sampleId) {
-    this.iterator++;
-    if(this.iterator === this.sampleHistory.length) {
-        // append to the history array if needed
-        this.sampleHistory.push(sampleId);
+    // only add a new item to the history if it was not just added
+    if(this.lastSample !== sampleId) {
+        // add new entry to array if no more space in current history
+        if(this.current === this.sampleHistory.length-1) {
+            this.current++;
+            this.sampleHistory.push(sampleId);
+        }
+        else {
+            // otherwise overwrite old history if room in array
+            this.current++;
+            this.sampleHistory[this.current] = sampleId;
+        }
+        this.lastSample = sampleId;
+        this.length = this.current+1;
     }
-    else {
-        // otherwise overwrite old history if room in array
-        this.sampleHistory[this.iterator] = sampleId;
-    }
-    // the length of the history will be equal to the iterator after adding
-    this.length = this.iterator;
 };
 
 /**
@@ -42,9 +47,9 @@ History.prototype.add = function(sampleId) {
  * @returns {string|null} The previous item in the list of samples.
  */
 History.prototype.back = function() {
-    if(this.iterator > 0) {
-        this.iterator--;
-        return this.sampleHistory[this.iterator];
+    if(this.current > 0) {
+        this.current--;
+        return this.sampleHistory[this.current];
     }
     else {
         return null;
@@ -59,9 +64,9 @@ History.prototype.back = function() {
  * @returns {string|null}
  */
 History.prototype.forward = function() {
-    if(this.iterator < this.sampleHistory.length - 1) {
-        this.iterator++;
-        return this.sampleHistory[this.iterator];
+    if(this.current > -1 && this.current < this.sampleHistory.length-1) {
+        this.current++;
+        return this.sampleHistory[this.current];
     }
     else {
         return null;
@@ -75,7 +80,7 @@ History.prototype.forward = function() {
  */
 History.prototype.reset = function() {
     this.sampleHistory = [];
-    this.iterator = 0;
+    this.current = -1;
     this.length = 0;
 };
 
--- a/public/js/History.js
+++ b/public/js/History.js
@@ -1,0 +1,81 @@
+/**
+ * History: Keep track of user's sample history.
+ *
+ * This object emulates the typical back/forward behaviour in browsers.
+ *
+ * @constructor
+ */
+function History() {
+    this.sampleHistory = [];
+    this.iterator = 0;
+    this.length = this.sampleHistory.length;
+    this.lastSample = null;
+}
+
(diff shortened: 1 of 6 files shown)


public/js/History.js
```diff
@@ -0,0 +1,81 @@
...
+ * @param {string} sampleId - The sampleID to add to the history.
+ */
+History.prototype.add = function(sampleId) {
+    if(sampleId !== this.lastSample) {
+        this.sampleHistory = this.sampleHistory.slice(0, this.iterator+1);
+        this.sampleHistory.push(sampleId);
+        this.iterator = this.sampleHistory.length - 1;
+        this.length = this.sampleHistory.length;
+        this.lastSample = sampleId;
+    }
```
jni: Shouldn't there be an else? Or is the if clause to prevent no-op clicks from adding to the history? If the latter, please add a comment to indicate this!

public/js/History.js
```diff
@@ -0,0 +1,81 @@
...
+ * @this {History}
+ * @param {string} sampleId - The sampleID to add to the history.
+ */
+History.prototype.add = function(sampleId) {
+    if(sampleId !== this.lastSample) {
+        this.sampleHistory = this.sampleHistory.slice(0, this.iterator+1);
+        this.sampleHistory.push(sampleId);
+        this.iterator = this.sampleHistory.length - 1;
+        this.length = this.sampleHistory.length;
+        this.lastSample = sampleId;
```
jni: Is this copying the full current history on every click? What is the data structure? If an array, it might be worth maintaining a pointer to the current iterator *and* to the current array size, so that we don't copy each time but rather overwrite existing, but obsolete, history.

public/js/script.js
```diff
@@ -99,6 +105,29 @@ function mountPlots(screenData, sampleData, featureNames) {
...
+    // attach behaviour to backwards and forwards buttons, unhide them
+    $backButton.unbind('click');
+    $forwardButton.unbind('click');
+
+    $backButton.on('click', function() {
+        var backId = history.back();
+        if(backId) {
+            lineplot.getSampleData(backId);
+            neighbourPlot.updatePoint(backId);
+            neighbourImages.getImages(backId);
```
jni: Hmm, another set of things that are all updated together. Perhaps worth having a more global function that updates the currently-selected sample globally? (ie in all models?) Soon the histogram will also need to be updated...

tests/History.spec.js
```diff
@@ -0,0 +1,60 @@
+var History = require('../public/js/History.js');
+
+describe('History', function() {
+    var history;
+
+    beforeEach(function() {
+        history = new History();
+    });
+
+    describe('add', function() {
```
jni: Nice! Relates to my comment about the if statement... Yeah a note in the docstring or under the if would be very useful to avoid confusion in the future!

jni: btw I love that you added tests! Super happy. =)

tests/History.spec.js
```diff
@@ -0,0 +1,60 @@
...
+    });
+
+    describe('reset', function() {
+        it('should remove all items from history', function() {
+            history.add('A');
+            history.add('B');
+            history.reset();
+            expect(history.length).toEqual(0);
+        });
+    });
```
jni: I think you need one more test: add A, add B, add C, back, add D, back, back (assert = B, A), forward, forward (assert = B, D). Ensures the C is not in the history still somehow.

public/js/History.js
starcalibre: It was there to stop the same sample being added twice in a row (same plot point being clicked).. but I've realised now it'd just be easier to remove the 'click' event from the active point on the graph. So this won't be necessary..!

public/js/script.js
starcalibre: I plan on completely re-factoring the main ``script.js`` file soon.. I'll be able to address this in that PR :)

tests/History.spec.js
starcalibre: Good idea, this test is going in!

public/js/History.js
starcalibre: It's returning a subset of the array, overwriting the old reference with new array, then the garbage collector will see there's no more reference to the old array and remove it. You raise a good point though, it's not the most efficient way to do it. JS Arrays are actually hash tables with sequential integer keys (remember everything in JS inherits from the base Object class, so this makes sense in topsy turvy JS land..). So the slice operation is O(n), and the push is O(1).. doing it the way you suggested will mean overwriting is O(1), and the push is O(1).. so O(1).. yay! :)

jni: > JS Arrays are actually hash tables with sequential integer keys

LOL WAT

Yeah, thinking about it, I guess this is actually true of Python lists, too...! Ugh.

starcalibre:
Should be good to go, thanks! :)

jni:
Erm, does the `on('click')` need to be re-added to that point once a user clicks away from it?

Also, I think `iterator` is the wrong name for the pointer, because iterators usually iterate over a collection once. What do you think of `this.current`? Or `this.present` as a pun on `history` if you're feeling whimsical. =)

starcalibre:
@jni SMH, if there was ever a case for why loose typing and coercion is a double edged sword.. this was it, all my tests were passing just fine, but the ``back`` function actually wasn't working as intended because my ``Array`` was being cast to an ``Object`` rather than throwing a fencepost error when the wrong value was being passed as the key! Ack! Sorry I left you Java, please take me back!!! 

Anyway, I found the bug and added a couple more tests to catch it along with added back the if statement as discussed! Is there a saying in unit tests "the one thing you don't test is the one thing that'll cause your program shit the bed in production?" because there should be!

jni:
@starcalibre Python 3.5 type annotations ftw! ;)

jni:
PS: thanks, you finally made me look up what "smh" meant. =P

public/js/History.js
```diff
@@ -0,0 +1,88 @@
...
+History.prototype.add = function(sampleId) {
+    // only add a new item to the history if it was not just added
+    if(this.lastSample !== sampleId) {
+        // add new entry to array if no more space in current history
+        if(this.current === this.sampleHistory.length-1) {
+            this.current++;
+            this.sampleHistory.push(sampleId);
+        }
+        else {
+            // otherwise overwrite old history if room in array
```
jni: For consistency, move this comment above the else statement

Other than that, great function, clear, compact, easy to follow! =)

Issue #49, opened 2015-05-07
Add loupe to image display


jni:
Now that we have thumb, large, and full image sizes, we should create a tool that lets users display the full size image on request, perhaps upon clicking on the large size.

starcalibre:
This could be implemented as a bootstrap modal (see [here](http://www.w3schools.com/bootstrap/bootstrap_modal.asp) for example). 

Underneath the image we could have information such as the sample ID, treatment, treatment target/gene name, etc.

starcalibre:
As discussed in today's meeting, it was suggested that clicking on the images should enable a 'compare samples' functionality. I still thinking zooming to the full-size image should be there. I'd rather not implement this as a right click context menu (this is a pain, and overriding the default behavior of the right click mouse button is counter-intuitive to how a user expects their browser to behave). My suggestions are a 'zoom' and 'add' button that are overlaid on the image when the mouse enters the image. See this quick mockup:

![untitled](https://cloud.githubusercontent.com/assets/7637964/7677327/71f5c362-fd8e-11e4-8549-03c1e7c00f5e.png)

jni:
Great writeup! I like your mockup and it's quite extensible to new functions!

starcalibre:
Idea for the functionality of the add button:

* Multiple samples can added plotted on the lineplot, each with a different colour that will correspond to a colour on the neighbour plot. 
* Add a tab under the large image in the image gallery that will let you select between images of the nearest neighbours, or images of the currently selected samples.

jni:
- Regarding lineplots, we could get very fancy here. I prefer to have the individual plots with low alpha/thickness, and an average plot (perhaps even with a stdev border) overlaid on top.
- I don't understand your second point?

PR #63, opened 2015-05-26
Image zoom in and add buttons

starcalibre:
This PR adds the zoom in and add buttons as discussed in #49. The zoom-in button works, but the add button is currently non-functional. The buttons appear only when the mouse hovers over the image. When the zoom button is clicked, a full-screen modal appears with the full-sized version of the image.
screenshot from 2015-05-26 14 04 42

### DIFF
--- a/public/js/NeighbourImages.js
+++ b/public/js/NeighbourImages.js
@@ -6,22 +6,26 @@ var _ = require('lodash');
  * @constructor
  */
 function NeighbourImages() {
+    var self = this;
     this.neighbours = [];
     this.selectedImage = "";
     this.neighbourImages = [];
 
-    // cache jquery selectors used frequency by gallery methods
-    // call these in the constructor so the DOM is traversed only once
-    // before caching.
+    // When a jQuery selector is called, the DOM must be traversed in order
+    // to find the element. if the selector is saved to a variable (ie cached), the
+    // element is remembered thus the DOM doesn't need to be traversed on subsequent
+    // calls. Frequently used selectors are cached in the object constructor so the
+    // selectors are cached when the object is created.
     this.selectors = {
         $nebula0: $('#nebula-0'),
         // create an array of jquery selectors for all the neighbour images
         $nebula: _.map(_.range(1, 24), function(value) {
-            return $('#nebula-' + (value))
+            return $('#nebula-' + (value));
         }),
         $body:  $('body'),
         $imgZoomButton: $('#img-zoom-button'),
-        $thumbnailChildImg: $('.thumbnail > img')
+        $thumbnailChildImg: $('.thumbnail > img'),
+        $imageButtons: $('#image-buttons')
     };
 
     var divWidth = Math.round($('#image-column').width());
@@ -37,17 +41,17 @@ function NeighbourImages() {
 
     // hide/unhide image buttons on hover
     $('#main-image').hover(function() {
-        $('#image-buttons').removeClass('hidden');
+        self.selectors.$imageButtons.removeClass('hidden');
     }, function() {
-        $('#image-buttons').addClass('hidden');
+        self.selectors.$imageButtons.addClass('hidden');
     });
 }
 
 /**
  * getImages: Get images for selected sample.
  *
- * Sends two AJAX requests to get the medium size image of the
- * inputed sample_id, and to get the images of the neighbouring
+ * Send two AJAX requests to get the medium size image of the
+ * input sample_id, and the images of the neighbouring
  * samples.
  *
  * @this {NeighbourImages}
@@ -58,12 +62,13 @@ NeighbourImages.prototype.getImages = function(sample_id) {
 
     var imageQueryLarge = {
         sample_id: sample_id,
-        select: ['image_large sample_id']
+        select: ['image_large sample_id'],
+        include: ['gene_name']
     };
 
     var imageQueryNeighbour = {
         sample_id: sample_id,
-        select: ['image_thumb sample_id'],
+        select: ['image_thumb sample_id gene_name'],
         neighbours: true
     };
 
@@ -82,7 +87,7 @@ NeighbourImages.prototype.getImages = function(sample_id) {
 
         })).then(function() {
 
-            // matches the height of the image gallery to the plots
+            // match the height of the image gallery to the plots..
             // do this *after* the images are loaded from the DB
             var height = Math.round($('#plot-column').height());
             $('#image-column')
@@ -103,7 +108,10 @@ NeighbourImages.prototype.getImages = function(sample_id) {
 NeighbourImages.prototype.setImages = function() {
     var self = this;
 
-    // make all img frames empty
+    // make all image frames empty, this is done to prevent
+    // images belonging to the previous sample selected
+    // from displaying if something goes wrong when the
+    // images are loaded.
     this.selectors.$nebula0
         .attr('src', '//:0')
         .attr('title', 'not found');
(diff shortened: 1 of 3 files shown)


public/js/NeighbourImages.js
```diff
@@ -6,14 +6,26 @@ var _ = require('lodash');
...
 function NeighbourImages() {
-    var self = this;
-
     this.neighbours = [];
     this.selectedImage = "";
     this.neighbourImages = [];
 
-    var divWidth = Math.round($('#image-column').width()) - 20;
-    var imgWidth = Math.round(divWidth/3) - 20;
+    // cache jquery selectors used frequency by gallery methods
```
jni: used frequently

public/js/NeighbourImages.js
```diff
@@ -6,14 +6,26 @@ var _ = require('lodash');
...
+    // call these in the constructor so the DOM is traversed only once
+    // before caching.
```
jni: There is the promise of greatness in this docstring! =) But it's still a bit off that: 

- is this just aliasing the names? Or what?
- what's the alternative design? (not in the constructor) I don't understand why this is more efficient than calling them on demand.

public/js/NeighbourImages.js
```diff
@@ -22,11 +34,21 @@ function NeighbourImages() {
...
+    });
 }
 
 /**
- * getImages: Get med-size image for selected samples, and nearest
- * neighbour images.
+ * getImages: Get images for selected sample.
+ *
+ * Sends two AJAX requests to get the medium size image of the
+ * inputed sample_id, and to get the images of the neighbouring
```
jni: inputed -> input

public/js/NeighbourImages.js
```diff
@@ -83,48 +102,83 @@ NeighbourImages.prototype.getImages = function(sample_id) {
...
+
+        this.selectors.$nebula[i]
             .attr('src', 'data:image/jpg;base64,' +
                 self.neighbourImages[i].image_thumb)
             .attr('title', self.neighbourImages[i].sample_id);
-        $nebulaSelector.unbind('click');
+        this.selectors.$nebula[i].unbind('click');
+
+        // use IIFE here as closure being called in a loop
+        // see http://www.mennovanslooten.nl/blog/post/62
```
jni: Wow. =) What does IIFE stand for?

public/js/NeighbourImages.js
```diff
@@ -83,48 +102,83 @@ NeighbourImages.prototype.getImages = function(sample_id) {
...
+NeighbourImages.prototype.openFullModal = function() {
+    var imageQuery = {
+        sample_id: this.selectedImage.sample_id,
+        select: ['image_full image_large sample_id']
+    };
+    $.ajax({
+        type: 'GET',
+        url: '/api/images/?' + $.param(imageQuery),
+        success: function(data) {
+            // TODO some collections don't have a fullsize image, use the large image in lieu
```
jni: Why is this to-do? Looks like it's done?

public/js/NeighbourImages.js
```diff
@@ -83,48 +102,83 @@ NeighbourImages.prototype.getImages = function(sample_id) {
...
+            var image = data[0].image_full || data[0].image_large;
+
+            // add image title
+            $('#full-size-img-title').html(data[0].sample_id);
```
jni: Do you have "treatment" in your mongos yet? (=P) If not, add gene_name to the title; otherwise, add treatment.

public/js/NeighbourImages.js
```diff
@@ -83,48 +102,83 @@ NeighbourImages.prototype.getImages = function(sample_id) {
...
     // make all img frames empty
-    $nebula0
+    this.selectors.$nebula0
         .attr('src', '//:0')
         .attr('title', 'not found');
 
-    $('.thumbnail > img')
+    this.selectors.$thumbnailChildImg
         .attr('src', '//:0')
         .attr('title', 'not found');
```
jni: Question: when you skip this step, do you end up "stacking" the images on top of one another, hogging resources? Or is this for something else I don't understand?

public/js/NeighbourImages.js
starcalibre: Whenever you use ``$('#my-element')``, you're basically telling jQuery "hey, go search the entire DOM for ``#my-element``. When you save the call to jQuery in a variable, you're saying "hey, search the DOM for ``#my-element`` then remember where it was so you don't need to search next time". 

Previously I was (dumbly) caching the selector in the methods instead of the constructor, so each time the method was called.. so it was searching the DOM every time the method was called. Now the selectors are object properties, they're cached for the entire life of the object!

Caching your jQuery selectors is a really good practice when you're making multiple calls to the same DOM element, and you don't expect the structure of the DOM to change. Which is exactly the perfect use case for the image gallery. :)

jni: BOOM. Please copy-paste this almost verbatim into the code! =) (You don't have to write you were being stupid. =P)

public/js/NeighbourImages.js
```diff
@@ -64,11 +77,17 @@ NeighbourImages.prototype.getImages = function(sample_id) {
...
             url: '/api/images?' + $.param(imageQueryNeighbour),
             success: function(json) {
-                json.shift();
                 self.neighbourImages = json;
             }
 
-        })).then(function(res, status) {
+        })).then(function() {
+
+            // matches the height of the image gallery to the plots
```
jni: matches -> match

public/js/NeighbourImages.js
```diff
@@ -22,11 +34,21 @@ function NeighbourImages() {
...
+        $('#image-buttons').addClass('hidden');
+    });
 }
 
 /**
- * getImages: Get med-size image for selected samples, and nearest
- * neighbour images.
+ * getImages: Get images for selected sample.
+ *
+ * Sends two AJAX requests to get the medium size image of the
```
jni: Sends -> Send

I think you see where this is going. =P

jni:
Good stuff! Gotta love all those browser-specific CSS flags. WTF happened to standards? =(

public/js/NeighbourImages.js
starcalibre: It's more sitting there so I don't forget to remove it once all the collections do have the fullsize image

public/js/NeighbourImages.js
starcalibre: Immediately Invoked Function Expression. Yes, self calling anonymous functions.. just when you thought it didn't get any better ;)

public/js/NeighbourImages.js
jni: Hmm. Ok that's fair. Maybe add a line: "remove once all collections reliably contain fullsize"

public/js/NeighbourImages.js
jni: Yeah definitely expand that for efficient googling in the future. =)

public/js/NeighbourImages.js
starcalibre: Cool, done.. apparently browsers can have new lines in ``title`` tags now, so the hover title can appear as:

MYORES-38038403-A01
GENE 

Which is neat :)

public/js/NeighbourImages.js
starcalibre: It's there in case images can't be found (be it because they're missing from the DB for some reason, or something went wrong during the AJAX request). I'd rather the frames in the image gallery be completely blank than still have neighbouring images belonging to the last sample that was looked at.

jni: Ah! Clever! You know the drill, can you add that explanation to the code, under "// make all img frames empty"? =)

public/js/NeighbourImages.js
starcalibre: I've added an issue for this!

This one will be a little more complicated! Currently the image gallery is pulling all of the images out of the images collection so it doesn't have access to the ``gene_name`` field. Rather than query the for this information again, I'm gonna re-think how data is shared is between each component of the app. I think it'll be smarter to move all the AJAX requests to the ``script.js`` in the event triggers so I can better manage what's being queried from the DB and where it's going.

jni:
Very nice, thank you for addressing the comments! =)

PR #65, opened 2015-06-04
Add option to switch between PCA and TNSE

starcalibre:
This PR adds the ability to choose between PCA and TNSE on the NeighbourPlot, and re-factors a few functions in the NeighbourPlot object so it's able to handle switching between the different spaces.

### DIFF
--- a/public/config/plots.js
+++ b/public/config/plots.js
@@ -1,0 +1,77 @@
+module.exports = {
+    neighbourPlot: {
+        // create space between the x and y axis. this value is a proportion
+        // of the range between the min and max. eg 0.02 creates a 2% margin
+        // between the points and x/y axis
+        axisMargin: 0.02,
+        // the duration of the transition between changes in the axis scale
+        // triggered by a change in dimensionality reduction methods
+        // this duration also applies to the transformation applied to the
+        // plot points
+        axisTransitionDuration: 750,
+        // the duration of the transition between active and inactive
+        // class for points
+        pointTransitionDuration: 150,
+        // the radius of the currently selected point on the plot
+        activePointRadius: 7,
+        // the radius of every other point, this should probably
+        // be smaller than the active point radius
+        inactivePointRadius: 5,
+        // the number of ticks to use in the x-axis, d3 will use
+        // this number of ticks, otherwise something close to it.
+        xAxisTicks: 10,
+        // as above, for the y-axis
+        yAxisTicks: 10,
+        // the margin between the edge of the SVG canvas and the
+        // edges of the 'drawing area' of the plot. this is needed
+        // so axis labels, ticks, ete. are rendered
+        margin: {top: 10, right: 40, bottom: 40, left: 40},
+        // at the time of drawing, the plots width is taken
+        // from its containing div. the height of the plot is calculated
+        // using this aspect ratio.
+        aspectRatio: {
+            width: 16,
+            height: 9
+        }
+    },
+    linePlot: {
+        // create space between the x and y axis. this value is a proportion
+        // of the range between the min and max. eg 0.02 creates a 2% margin
+        // between the points and x/y axis
+        axisMargin: 0.02,
+        // the number of ticks to use in the x-axis, d3 will use
+        // this number of ticks, otherwise something close to it.
+        xAxisTicks: 8,
+        // as above, for the y-axis
+        yAxisTicks: 5,
+        // the margin between the edge of the SVG canvas and the
+        // edges of the 'drawing area' of the plot. this is needed
+        // so axis labels, ticks, etc. have enough room to draw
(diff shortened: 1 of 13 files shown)


public/index.html
```diff
@@ -58,6 +58,15 @@
     <div class="container">
         <div class="row">
             <div class="col-lg-8" id="plot-column">
+                <div id="neighbourplot-options" class="row hidden">
+                    <label for="dimensionality-reduction-select">
+                        Dimensionality Reduction:
+                    </label>
+                    <select id="dimensionality-reduction-select" class="form-control">
+                        <option value="tnse">TSNE</option>
```
jni: LOL tnse -> tsne

Where is the mongo query for this? I'm updating `microscopium` to create the tSNE features, so I want to make sure the schema matches.

starcalibre:
@jni The typo is now fixed and updated schema has been added to the readme

public/js/NeighbourPlot.js
```diff
@@ -66,109 +66,218 @@ function NeighbourPlot(sampleData, element) {
...
+
+
+    // define tooltip function and append to SVG
+    if(!this.tip) {
+        this.tip = d3.tip()
+            .attr('class', 'd3-tip')
+            .offset([-10, 0])
+            .html(function(d) {
+                return "<p><strong>ID: </strong>" + d._id + "</span></p>" +
+                    "<p><strong>Gene: </strong>" + d.gene_name + "</span></p>";
```
jni: Should this not be defined by content classes and CSS styles?

jni: Also, do you want to roll the new naming scheme ("treatment") here?

public/js/NeighbourPlot.js
```diff
@@ -66,109 +66,218 @@ function NeighbourPlot(sampleData, element) {
...
-        .call(self.tip);
-
-    // append axis
-    self.svg.append('g')
-        .attr('class', 'x axis')
-        .attr('transform', 'translate(0,' + self.height + ')')
+    // apply the axis to the containers, an animation is applied when they're being updated
+    this.xAxisSvg
+        .transition()
+        .duration(750)
```
jni: Unhappy with this magic number. Can we make a config file where all these constants are read from? e.g. .yaml. At a minimum it should be defined at the start of this function.

jni:
Good-o. =) Very minor comments.

Overall I think the codebase is getting a bit complex, but we can leave a cleanup for a bit longer still.

public/js/NeighbourPlot.js
starcalibre: I like this idea! Maybe it could be just be a JSON file where each key has the settings for some part of the UI.

That way plot behaviour/appearance can be tweaked quite easily from one location, or even adjusted by users if that's something we want to do much later down the track.

public/js/NeighbourPlot.js
starcalibre: If I wanted to do anything other than bold it, sure.. but now I think about it I'm not a big fan of semantic tags so I think I'll use CSS here to be consistent instead. And move this into its own function while I'm at it. 

Sure -- the MYORES and VCFG datasets currently don't have that field, but I'll just add it now.

starcalibre: Ohh wait, the bold text is already done in the CSS! I'll just remove the redundant markup in the tooltip :)

starcalibre:
@jni Okay, this one should be good to go now!

public/config/plots.js
```diff
@@ -0,0 +1,77 @@
+module.exports = {
+    neighbourPlot: {
+        // create space between the x and y axis. this value is a pct of the
+        // range of the data plotted to the axis. eg 0.02 creates a 2% margin
+        // between the points and x/y axis
```
jni: Man, I love these detailed comments. More of this, please! =)

public/index.html
```diff
@@ -58,6 +58,15 @@
     <div class="container">
         <div class="row">
             <div class="col-lg-8" id="plot-column">
+                <div id="neighbourplot-options" class="row hidden">
```
jni: Is this still a pull-down? It occurs to me that we could just make it a button. I would also rename it to "scatterplot view" or something like that, a bit friendlier to users. I can't find an exact example of what I mean right now but something along the lines of the top row of this image: https://d3ui957tjb5bqd.cloudfront.net/images/screenshots/products/15/152/152410/creativemarket-f.jpg

starcalibre:
scatterplot-view

 @jni there we go! what do you think?

jni:
I think it's gorgeous! Can you just update 'View' -> 'view'?

As an aside, not the best example of tSNE's performance. =\

jni:
Yay! =) Thanks, really awesome work!

PR #69, opened 2015-09-24
Refactor lineplot object

starcalibre:
The first of the promised series of PR.. this one is an example of how I've been re-writing the plot objects. Once this one is merged, I can send PRs for the rest of the plots then start adding some new features to the UI (excite!).

I've mostly just broken up lots of the larger functions into smaller private methods. Note: JS has no true private methods, but it's convention to use the ``_`` suffix for any method to be treated as private. (same as Python, right?)


public/js/Lineplot.js
```diff
@@ -209,56 +216,90 @@ Lineplot.prototype.updateActiveLine = function(feature) {
...
+
+/**
+ * updateActiveLine: Redraw the line overlaying selected feature in the plot.
  *
  * @this {Lineplot}
+ * @private
  */
-Lineplot.prototype.destroy = function() {
-    d3.select(this.element + ' > svg').remove();
+Lineplot.prototype._updateActiveLine = function() {
```
jni: "Lineplot" is a very generic name, whereas this function is for a very specific line in a very specific plot. So there's some API tweaking to be done here...

starcalibre: ``FeatureVectorLinepot`` and and the equally creative ``FeatureDistributionHistogram`` for the histogram ?

jni: Yes these two names sound appropriate. =) Well, `LinePLOT`, not `LinePOT`. =P

public/js/script.js
```diff
@@ -117,7 +117,7 @@ function mountPlots(screenData, sampleData, featureNames) {
 
     var neighbourImages = new NeighbourImages();
     var histogram = new Histogram(screenData._id, featureNames, '#histplot');
-    var lineplot = new Lineplot(screenData._id, '#lineplot');
+    var lineplot = new Lineplot('#lineplot');
```
jni: Same here re:naming. Make sure to name variables as meaningfully as possible.

public/js/Lineplot.js
jni: Oh and again, in this context, "updateActiveFeatureVline", for example

public/js/Lineplot.js
```diff
@@ -209,56 +216,90 @@ Lineplot.prototype.updateActiveLine = function(feature) {
...
+    var xCoord = this.xScale(this.activeFeature);
+    this.svg.selectAll('.selectedLine').remove();
+    this.svg.append('line')
+        .attr('x1', xCoord)
+        .attr('y1', 0)
+        .attr('x2', xCoord)
+        .attr('y2', this.height)
+        .attr('stroke', 'red')
+        .attr('stroke-width', 0.75)
```
jni: magic numbers are... baaaaaad.

public/js/Lineplot.js
```diff
@@ -209,56 +216,90 @@ Lineplot.prototype.updateActiveLine = function(feature) {
...
+        .attr('stroke-dasharray', '5, 5')
```
jni: ditto. What's with the string type, btw?

public/js/Lineplot.js
starcalibre: I should probably define these styles in the CSS anyway!

public/js/Lineplot.js
```diff
@@ -39,24 +38,25 @@ function Lineplot(screen, element) {
...
     var aspectHeight = config.aspectRatio.height;
     this.fullWidth = $(this.element).width();
-    this.fullHeight = Math.round(this.fullWidth * (aspectHeight/aspectWidth));
+    this.fullHeight = Math.round(this.fullWidth *
+        (aspectHeight / aspectWidth));
 
-    this.width = this.fullWidth - this.margin.left - this.margin.right;
-    this.height = this.fullHeight - this.margin.top - this.margin.bottom;
+    this.width = this.fullWidth - this.margin.left -
+        this.margin.right;
```
jni: Why is this on two lines? It's less than 80c...

public/js/Lineplot.js
```diff
@@ -39,24 +38,25 @@ function Lineplot(screen, element) {
     var aspectWidth = config.aspectRatio.width;
     var aspectHeight = config.aspectRatio.height;
     this.fullWidth = $(this.element).width();
-    this.fullHeight = Math.round(this.fullWidth * (aspectHeight/aspectWidth));
+    this.fullHeight = Math.round(this.fullWidth *
+        (aspectHeight / aspectWidth));
```
jni: This line break seems unnecessary as well

public/js/Lineplot.js
starcalibre: It's an SVG attribute, they always take strings as their type. Numbers just end up getting parsed as strings anyway :)

eg.
```
<svg height="210" width="500">
  <line x1="0" y1="0" x2="200" y2="200" style="stroke:rgb(255,0,0);stroke-width:2" />
</svg>
```

public/js/Lineplot.js
```diff
@@ -65,139 +65,146 @@ Lineplot.prototype.getSampleData = function(sampleId) {
...
         .call(yAxis);
+};
+
+/**
+ * drawLine: Draw the line for the lineplot.
+ *
+ * @private
+ */
+Lineplot.prototype._drawLine = function() {
+    var linePoints = _.zip(_.range(1, this.featureVector.length + 1),
```
jni: Note here why you're using range(1, n+1)

public/js/Lineplot.js
starcalibre: 81 characters without the break.. :C

jni:
You are in like Flynn! =P

PR #70, opened 2015-10-05
Refactor Histogram object

starcalibre:
This is similar to the refactor of the Lineplot. The Histogram has been renamed to ``FeatureDistributionHistogram`` and larger functions have been broken up into smaller private functions


jni:
Nice! I think your sense of code aesthetics is improving. =D

PR #72, opened 2015-11-12
Add UI controller object

starcalibre:
This PR adds a UI controller object. This object handles all of the UI/plot updates whenever something is clicked. This should make the code easier to maintain and extend with new features.

There's also a couple of minor additions:
* An enum for keyboard event keycodes
* Update to the latest ``lodash`` -- there's been a few optimizations added in the last few release, so best take advantage of these!

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,6 @@
 {
   "name": "microscopium-ui",
   "main": "server.js",
-  "scripts": {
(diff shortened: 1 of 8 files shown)


jni:
Nice, clean PR, @starcalibre!

Issue #47, opened 2015-04-29
Improve initial loading screen


jni:
It currently takes just a little bit too long to load a dataset once a screen has been selected (or on page load when there is a single screen). It *only just* gets to the point where you wonder if the site is broken. Possible options:

- include a progress bar
- save the initial plot state somewhere (eg as part of the screen info) and display that.
- optionally, add each data point to the plot as it is processed.
- something else I haven't thought of.

starcalibre:
* Pre-compute the bins needed to render the histogram and save these in the screen document, and fetch the feature vector for each sample from the DB each time a sample is clicked. This will greatly reduce the amount of data sent when the screen is loaded.
* Use the maximum level of compression possible when sending the data during the initial screen load

starcalibre:
#51 makes a great improvement to the initial load time.

starcalibre:
The Flask [docs](http://flask.pocoo.org/docs/0.10/patterns/streaming/) show an example of streaming a CSV line-by-line using a generator. This pattern could be adapted to Mongo cursors to stream data during loading.

PR #75, opened 2015-12-02
Use server-side templating to render HTML

starcalibre:
As discussed Monday, this PR adds the server-side templating using ``Jinja2``. 

* Larger HTML files have been broken down into smaller files
* Screens are now presented in a table instead of a dropdown menu
* Some JS was updated to play nicely with the new templating
* HTML will be minified when the UI is started in "production" mode.


starcalibre:
@jni this should be good to go :)

jni:
@starcalibre added a few more comments. Also, can you comment on the feasibility of live-loading a dataset (without spinners)? (Obvs doesn't have to happen now, just want to know if it's possible, and if so, add an issue so that we remember to do it at some point.)

starcalibre:
@jni I'm sure it's possible. The flask docs give an [example](http://flask.pocoo.org/docs/0.10/patterns/streaming/) of using a generator to send a CSV line-by-line as a series of responses. I don't see why such a pattern couldn't be adapted to a Mongo cursor.. 

We already have issue #47 for improving the initial load times and process, I'll add a comment there. :)

jni:
@starcalibre LOL, it's already one of the points in that issue. =)

starcalibre:
@jni Happy new year! Should be good to go now..!

jni:
And a very happy new year to you, sir! =)

PR #76, opened 2016-01-05
Add byteflag utility class

starcalibre:
The new fast canvas based scatterplot I've been working on has involved the creation of a few new functions and classes, so I thought I'd merge these in iteratively to make the process a bit easier.

One big disadvantage to using the canvas to draw plots is you can't save state in the DOM -- the current scatterplot manages details such as which is the active sample, which is being filtered out, etc using CSS classes. You don't get this convenience with the canvas, so you need your own way to manage the state of each element being drawn.

JS supports C-style bit manipulation, so I've been managing the state of each sample using a byteflag. A ``Number`` is attached to every sample and is treated as a byteflag. e.g. samples with value ``0x01`` are the actively selected point, samples with ``0x02`` are neighbour points, etc. Whenever the plot drawing method is called, the current state of each sample can be read from the byteflag so the point can be drawn with the appropriate stylings. I thought this was the best way to go because it elegantly handles multiple states for each point while only using a single ``Number`` to do so.

Bit manipulation syntax is dog ugly and easy to make mistakes with so this PR ads a few methods to make bit manipulation easier. It also moves all the utility classes to a ``utils`` folder. Tests for great justice included too. :D


starcalibre:
@jni Just checking you got the notification for this one? :)

jni:
@starcalibre I did but been stupid busy with book and such... Will look later today... Maybe ping me again though... =)

app/static/js/utils/Byteflag.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+var Byteflag = {};
+
+/**
+ * add: Add bit to byteflag.
+ *
+ * @param flag {number} - Query byteflag.
+ * @param newBit {number} - New bit too add.
```
jni: too -> to

app/static/js/utils/Byteflag.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+var Byteflag = {};
+
+/**
+ * add: Add bit to byteflag.
```
jni: This description makes it seem like it's done in-place. Maybe something along the line, "Return new byteflag with newBit added to existing flag"

tests/Byteflag.spec.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
```
jni: ?

tests/Byteflag.spec.js
```diff
@@ -0,0 +1,46 @@
...
+            expect(check).toBe(true);
+        });
+
+        it('should return true when query bit not present', function() {
+            var flag = (0x01 | 0x04);
+            var check = Byteflag.check(flag, 0x02);
+            expect(check).toBe(false);
+        });
+    });
+});
```
jni: Nice, elegant tests! =) The best code is when it feels like you're reading English. =)

jni:
@starcalibre some very minor comments. =)

tests/Byteflag.spec.js
starcalibre: It turns on JavaScript's [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) -- it changes the behaviour slightly to prevent oopsies like declaring global variables, and throws errors instead of failing silently with certain actions. It's good practice to always use it and I plan on adding it to the existing code at some point in the near future.

starcalibre:
@jni thanks, should be good to go now !

tests/Byteflag.spec.js
jni: I figured it would be something like this, but I resent the fact that it's a string literal. =)

PR #77, opened 2016-01-25
Add new version of Scatterplot

starcalibre:
Okay, here it is.. finally.. the new Canvas based scatterplot with panning, zooming, and scalability to large datasets.

The idea behind the plot is that it creates multiple layers of SVG and Canvas drawing contexts. Some elements are drawn to the SVG (axis, tooltips) while points are drawn using the canvas. This gives you best of both worlds -- SVG rendering for text elements, d3 axis functionality, etc. while letting the much faster canvas do the heavy lifting with drawing all the points in the plot. The functionality of the new plot is managed with the ``NeighbourPlotCanvas`` object, which replaces the previous ``NeighbourPlot`` object.

The downside to the Canvas is you need to manage the state of each object you're drawing yourself. There's no convenience of CSS styles for this. As I mentioned in an earlier PR, I'm using a byteflag that gets assigned to each sample to do this.``SampleManager`` wraps the current collection of Sample objects, and manages these byteflags.

All the drawing is handled by the ``PointsDrawer`` object. This object draws the points in the style corresponding to their status byteflag, and in the correct order, i.e. the actively selected point and its neighbours should be drawn last so they're visible on the plot.

I've marked this as WIP because I need to add some more comments to the ``NeighbourPlotCanvas`` object as you'll probably notice, but everything else is good for review.


jni:
@starcalibre very neat! Other than the "Add NeighbourPlotCanvas object" update, clean and readable work. That particular commit needs some major work though...

starcalibre:
@jni Thanks for taking the time to add those comments! Yeah a few of the comments and docstrings aren't complete in ``NeighbourPlotCanvas`` (I had it marked as WIP for that reason), but I know where the confusing parts are so should be able to get it finished fairly easily.

starcalibre:
@jni Got a bit sidetracked with linuxconf and the other things I've been working on, but this should be good to go now!

jni:
And so it is! =)

Issue #39, opened 2015-03-12
Overlay screen single-z-score result onto the neighbour plot


jni:
Every screen we get has a z-score result, or a list of hits. It would be great to be able to see these on the neighbour plot / dimension-reduced scatterplot.

starcalibre:
It seems like there's a few different values that could be overlayed onto the scatterplot.. z-score/hits, viability, a PCA value, etc. I suggest adding a radio button menu that will change the colouring of the points according to the chosen option. The radio buttons could be generated according to what data is available for the screen.

jni:
I think at this point I want to simplify before we go wild on the options. So, let's assume z-score initially, and add other options as users request them.

jni:
Use a colormap to display the value. Matplotlib's new [viridis colormap](http://bids.github.io/colormap/) has been implemented in D3 (see links in above page). I suggest we use that. (A light-gray background might be needed for that, though.)

starcalibre:
Here's a a few [examples](http://bl.ocks.org/starcalibre/6cccfa843ed254aa0a0d) of some colour scales -- I think viridis looks good, even without a light gray background.

Here's how I think I should add this..

Add a new field to the ``Sample`` document, call it ``available_overlays``. In this field store an array an strings of which information is available for the screen. Some screens have nothing, while other's like Keefe's are multi-parametric so there's a few different scores to overlay. The contents of this array are used to generate the drop-down menu to change which number is used to generate the colour scale. If there's no additional information for the overlays, the dropdown menu will be empty or won't render at all.

Each sample then has the scores saved in an object, ``overlays``. So let's say I have two scores.. ``z_score`` and ``viability``. The ``Screen`` document will contain two strings ``z_score`` and ``viability``. The ``Sample`` document field object will be ``{ z_score: -1.45, viability: 2.3 }``.

jni:
Hey,

I love both viridis and PuOr! I think it does depend on the application whether one prefers a diverging or a single colorscale, so make it at least configurable at start-up time.

Your implementation idea sounds reasonable. Do you think it would be worthwhile to add a column-wise version of the storage for these attributes? It might affect performance quite dramatically.

starcalibre:
It could even be configurable during run time, as per the example I just posted! :)

As long as the order of the ``Sample`` documents in the array delivered from Mongo is guaranteed to match the order of column-wise storage, that can be done. I'm not sure what the performance benefit would be, if any.

starcalibre:
[...]

PR #78, opened 2016-02-16
Add point colour overlays for displaying z-scores (or other info)

starcalibre:
This PR adds the functionality to the UI to display z-scores (or other numeric information such as cell viability).

* On the database side of things, I've used the approach I described in #39 . An ``available_overlays`` array is populated in the ``Screen`` document that describes which overlays are available for the screen. If there's overlays available, the Flask template renders the dropdown menu. No menu is rendered if no screen overlay info is available so the UI just behaves as it did before. :)
* I was going to add support for Date/Time information, but all the images for Keefe's screen were either all done on the same date or the same date just got added to each plate. I think it'll be worthwhile to have this once we have a screen where date information is available though.
* I added two functions for calculating percentiles but didn't end up using them in the end. They're elegant af and could end up being useful. Don't mind removing them if you're not keen on unused code though!
* The upper and lower limits for the score colour scale is defined as config variable that forces the range [-4, 4]. Any values above or below that will be clamped to the min/max respectively. I wasn't quite sure how to handle this. I was going to use the 5th and 95th percentiles to calculate the upper and lower limits but it made the diverging scales look like ass because 0 wouldn't map to the "middle" neutral colour. Do you have any suggestions?
* Colour scales supported are magma, inferno, viridis and plasma from the new matplotlib scales.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,9 @@ Prerequisites
 * NPM package manager
 * Bower package manager
 * MongoDB version > 2.6
-* A modern browser, at least IE10+ or equivalent.
+* A modern browser. The interface is tested as working on 
+    latest Chrome, Firefox, Edge and IE11. Due to the heavy use of the HTML5Canvas API,
+    performance is best on Chrome.
 
 Install Instructions (Development)
 ----------------------------------
@@ -71,9 +73,14 @@ following format:
 var Screen = {
     _id: String, // A short ID for the screen.
     screen_desc: String, // A description of the screen.
-    number_samples: Number // The number of
-    screen_features: Array // An array of strings labelling each feature used in
+    number_samples: Number, // The number of
+    screen_features: Array, // An array of strings labelling each feature used in
                            // the screen.
+    available_overlays: Array // An optional field. An array of additional overlay
+                              // information available for samples in this screen.
+                              // e.g. if samples have ``screen_run`` and ``z_score``
+                              // overlay information, this array should be 
+                              // ["screen_run", "z_score"]
 }
 ```
 
@@ -89,7 +96,7 @@ var Sample = {
                             // applied to the sample.
         feature_vector: Array,  // An array of Numbers, the vector of
                                 // unstandardised features
-        featute_vector_std: Array  // An array of Numbers, the vector of standardised
+        feature_vector_std: Array  // An array of Numbers, the vector of standardised
                                    // features.
         neighbours: Array  // An array of sample_id strings, the 25 nearest
                            // neighbours of the sample.
@@ -101,10 +108,11 @@ var Sample = {
             pca: Array // the co-ordinates for the sample's point after PCA dimension reduction
             tsne: Array // the co-ordinates for the sample's point after TSNE dimension reduction
         },
-        image: ObjectID, // image_id  The ID of the samples image document
-                         // (see below).
-        cluster_member: Array // Array of integers showing cluster memberships
-                              // for clustering scheme k=2..20
+        overlays: Object // An object of the additional screen overlay information
+                         // for this sample. This information is optional. If no 
+                         // overlay information is available, the corresponding
+                         // sample document should no, or an empty, ``available_overlays``
+                         // field. Currently supports numeric and Date data.
     }
 ```
 
--- a/app/static/config/plots.js
+++ b/app/static/config/plots.js
@@ -69,7 +69,20 @@ module.exports = {
                 radius: 5
             }
         },
-        scaleExtent: [1, 10]
+        // the minimum/maxium scale factor allowed when zooming into the scatterplot
+        zoomExtent: [1, 12],
+        // the colourscale to use when plotting overlays to the neighbourplot
+        // should be one of the new matplot scales: inferno, magma, plasma or viridis
+        // see: https://bids.github.io/colormap/
+        colourScale: 'viridis',
+        // the minimum and maximum z scores to display on the colourmap
+        // outside the range will be clamped
+        colourScaleExtent: [-4, 4]
+    },
+    neighbourPlotLegend: {
+        width: 230,
+        height: 50,
+        margin: {top: 5, bottom: 20, left: 5, right: 5}
     },
     featureVectorlinePlot: {
         // create space between the x and y axis. this value is a pct of the
--- a/app/static/config/plots.js
+++ b/app/static/config/plots.js
@@ -77,7 +77,7 @@ module.exports = {
         colourScale: 'viridis',
         // the minimum and maximum z scores to display on the colourmap
         // outside the range will be clamped
-        colourScaleExtent: [-4, 4]
+        colourScaleExtent: [-3, 3]
     },
     neighbourPlotLegend: {
         width: 230,
--- a/app/static/js/NeighbourPlotCanvas.js
+++ b/app/static/js/NeighbourPlotCanvas.js
@@ -1,6 +1,37 @@
 var _ = require('lodash');
 var d3 = require('d3');
+
 var config = require('../config/plots').neighbourPlot;
+
+// browserify can't load modules dynamically due to the way it statically
+// analyses paths when creating the output bundle. in other words, no variables
+// or concatenated strings can be used in required paths.
+// this switch is inelegant and should be replaced with something nicer
+// maybe this could be configured server-side maybe when matplotlib 1.5 is
+// released and these colour scales become available??
+// TODO replace this whole godfawful switch
+var colourMap;
+switch(config.colourScale) {
+    case 'inferno':
+        colourMap = require('scale-color-perceptual/hex/inferno');
+        break;
+    case 'magma':
+        colourMap = require('scale-color-perceptual/hex/magma');
+        break;
+    case 'plasma':
+        colourMap = require('scale-color-perceptual/hex/plasma');
+        break;
+    case 'viridis':
+        colourMap = require('scale-color-perceptual/hex/viridis');
+        break;
+    default:
+        colourMap = require('scale-color-perceptual/hex/viridis');
+        break;
+}
+
+var Utils = require('./utils/Utils.js');
+var legendConfig = require('../config/plots').neighbourPlotLegend;
+var PlotLegend = require('./PlotLegend.js');
 var PointsDrawer = require('./PointsDrawer.js');
 var SampleManager = require('./SampleManager.js');
 var status = require('./enums/sampleStatus.js');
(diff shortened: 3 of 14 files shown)


jni:
@starcalibre can you add some screenshots to this PR?

btw, you mention diverging colorscales but all your supported scales are linear...

Regarding how to deal with the problem of mapping diverging scales so that 0 is in the centre, just map it from [-max(abs(percentile5, percentile95)), +max(abs(percentile5, percentile95))]. =)

starcalibre:
@jni I was using a few diverging scales but ended up using the sequential matplotlib scales in the end due to the problem I was having with the diverging scales (also good to have them because they're really nice). I'll try your suggestion and see how that goes!

I'll add screenshots and a video shortly!

starcalibre:
screenshot from 2016-02-16 16 32 32

starcalibre:
@jni  https://www.dropbox.com/s/v1dmerjrbnai160/out.ogv?dl=0 - video link. it looks awful and overly compressed when I stream it from dropbox, so you might need to download it if you have that problem

README.md
```diff
@@ -9,7 +9,9 @@ Prerequisites
 * NPM package manager
 * Bower package manager
 * MongoDB version > 2.6
-* A modern browser, at least IE10+ or equivalent.
```
jni: Yeah wtf, how did this get past review? =P

app/static/config/plots.js
```diff
@@ -69,7 +69,20 @@ module.exports = {
...
-        scaleExtent: [1, 10]
+        // the minimum/maxium scale factor allowed when zooming into the scatterplot
+        zoomExtent: [1, 12],
+        // the colourscale to use when plotting overlays to the neighbourplot
+        // should be one of the new matplot scales: inferno, magma, plasma or viridis
+        // see: https://bids.github.io/colormap/
+        colourScale: 'viridis',
+        // the minimum and maximum z scores to display on the colourmap
+        // outside the range will be clamped
+        colourScaleExtent: [-4, 4]
```
jni: I would say make this -3, 3. Anything more doesn't make sense.

app/static/js/NeighbourPlotCanvas.js
```diff
@@ -1,6 +1,37 @@
...
 var d3 = require('d3');
+
 var config = require('../config/plots').neighbourPlot;
+
+// browserify can't load modules dynamically due to the way it statically
+// analyses paths when creating the output bundle. in other words, no variables
+// or concatenated strings can be used in required paths.
+// this switch is inelegant and should be replaced with something nicer
+// maybe this could be configured server-side maybe when matplotlib 1.5 is
+// released and these colour scales become available??
```
jni: Huh? mpl 1.5 is available both in PyPI and conda. But how does this help in js? At any rate I don't find the switch statement *that* offensive!

app/static/js/NeighbourPlotCanvas.js
```diff
@@ -111,6 +146,30 @@ NeighbourPlotCanvas.prototype.setScale = function() {
...
+        colourScale = null;
+    }
+    else {
+        colourScale = d3.scale.linear()
+            .domain(Utils.linspace(scaleMin, scaleMax, colourMap.length))
+            .range(colourMap)
+            .clamp(true);
+    }
+
+    // new new colourscale to pointsdrawer
```
jni: new new

app/static/js/NeighbourPlotCanvas.js
```diff
@@ -257,6 +316,15 @@ NeighbourPlotCanvas.prototype._createCanvasSVGSelectors = function() {
         // mouse events
         .style('z-index', 3)
         .style('position', 'absolute');
+
+    // SVG for drawing legend
+    this.legendSvg = parentDiv.append('svg')
+        .attr('width', legendConfig.width)
+        .attr('height', legendConfig.height)
+        .style('transform', 'translate(' + (this.margin.left + 1 + this.width
+            - legendConfig.width - 5) + 'px' + ',' + (this.margin.top + 5) + 'px)')
```
jni: This is messy... What's happening and can it be cleaned up?

app/static/js/NeighbourPlotCanvas.js
```diff
@@ -274,7 +342,7 @@ NeighbourPlotCanvas.prototype._createNavigateBehaviour = function() {
     return d3.behavior.zoom()
         .x(this.xScale)
         .y(this.yScale)
-        .scaleExtent(config.scaleExtent)
+        .scaleExtent(config.zoomExtent)
```
jni: Is this an error from a previous PR? If so, is it feasible to write a test that catches it?

starcalibre: No, I renamed it from scaleExtent to zoomExtent in the config. Less
ambiguity!

app/static/js/PlotLegend.js
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * setColourScale: Draw the legend for the given colour scale.
+ *
+ * @param scale {Function} - Instance of a d3.scale.linear() where the range
+ *     is an array of colours. (i.e a colour scale).
+ */
+PlotLegend.prototype.setColourScale = function(scale) {
+    this.svg.selectAll('*').remove();
+
+
```
jni: Two blank lines seems excessive here?

app/static/js/NeighbourPlotCanvas.js
jni: Ah. Whew! =)

app/static/js/NeighbourPlotCanvas.js
starcalibre: My idea was to generate the scale (which is just an array of RGB strings)
using matplotlib or Palettable or something on the sever and send that
along with the template, rather than loading up a whole bunch of them in
JS. I'm probably over thinking it though..

app/static/js/PointsDrawer.js
```diff
@@ -13,6 +14,17 @@ function PointsDrawer(canvas) {
     this.width = canvas.node().width;
     this.height = canvas.node().height;
     this.context = canvas.node().getContext('2d');
+
+    this.overlay = 'None';
+
+    // default colourscale
+    // this colourscale is used when no overlay is active
+    this.defaultColourScale = d3.scale.linear()
+        .domain([-3, 3])
```
jni: -3, 3?

app/static/js/PointsDrawer.js
```diff
@@ -75,11 +87,21 @@ PointsDrawer.prototype.redraw = function (data, indices) {
     }
 
     if(neighbours.length > 0) {
-        this._drawPoints(data, neighbours, config.pointStyle.neighbours);
+        this._drawPoints(data, neighbours, config.pointStyle.neighbours, true);
     }
 
     if(!_.isNull(active) && !_.isUndefined(active)) {
-        this._drawPoints(data, [active], config.pointStyle.active);
+        this._drawPoints(data, [active], config.pointStyle.active, true);
```
jni: What is `true` here and can you add a comment to clarify?

app/static/config/plots.js
starcalibre: Yeah that makes sense.. but I'll try and implement your +/- abs(max(5th
percentile, 95th percentile)) idea first?

app/static/js/NeighbourPlotCanvas.js
jni: I think you are. Does D3 do linear interpolation between these automatically?

app/static/js/utils/Utils.js
```diff
@@ -88,4 +88,67 @@ Utils.euclideanDistance = function(x1, y1, x2, y2) {
...
+ * @param end {number} - The last value.
+ * @param n - {number}
+ * @returns {Array} - The linearly spaced array.
+ */
+Utils.linspace = function(start, end, n) {
+    var out = [];
+    var delta = (end - start) / (n - 1);
+
+    var i = 0;
+    while(i < (n - 1)) {
```
jni: Doesn't js have a for-loop??

app/static/js/utils/Utils.js
```diff
@@ -88,4 +88,67 @@ Utils.euclideanDistance = function(x1, y1, x2, y2) {
...
+ * Given an array of numbers, find the given percentiles. E.g
+ * getPercentiles(queryArray, [0.25, 0.5, 0.75]) will find the 25th,
+ * median and 75th percentile values.
+ *
+ * @param data - {Array} - The query Array. Does not need to be sorted.
+ * @param percentiles {Array} - An array of percentile values to find.
+ * @returns {Array} - The array of percentile values.
+ */
+Utils.getPercentiles = function(data, percentiles) {
+    var sorted = data.sort(function(a, b) { return a - b });
```
jni: Does sort really not have a default comparison for numbers???

jni:
Pretty good! Only minor comments. =)

app/static/config/plots.js
jni: oh, sure. =) that'll actually translate to something closer to [-1.9, 1.9] =)

app/static/js/utils/Utils.js
starcalibre: it does, but the default casts everything to a string during the
comparison. So [1, 20, 2, 10] will be sorted to [1, 10, 2, 20] instead of
[1, 2, 10, 20].

app/static/js/NeighbourPlotCanvas.js
starcalibre: Yeah, the d3 linear scale always interpolates.

app/static/js/PointsDrawer.js
starcalibre: Ahh that call to domain doesn't need to be there actually! The scale will map to the default colour whatever you feed into it..

app/static/js/utils/Utils.js
jni: WAT

starcalibre:
@jni Take 2!

I've decided to set the z-score range to ``[-3, 3]``. The percentiles idea was great in theory, but it kept slightly changing the axis on the legend time each time z-score used for the overlay was changed!! Really confusing. I think percentiles might be useful for other types of numbers (such as cell viability) so I'll keep the percentile functions in even if I'm not using them yet.

jni:
@starcalibre I added some very minor comments — third time's the charm! =D

starcalibre:
@jni should be good to go now:)

jni:
@starcalibre and so it is. =D

PR #79, opened 2016-03-03
Add new .jshintrc file and enable strict mode all JS files

starcalibre:
Add new .jshintrc file and enable strict mode all JS files

This PR adds a stricter ``.jshintrc`` file based on the [Airbnb JS style guide](https://github.com/airbnb/javascript/tree/master/es5) with a few small changes.

* [Strict mode](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Strict_mode) is enabled for all JS files, and will now be enforced by JSHint. Strict mode prevents whoopsies that usually fail silently in JS such as using undeclared variabled and using duplicate property names in functions.
* The stricter ``.jshintrc`` file caught a few things like missing semi-colons, lines longer than
80 characters, etc. Pretty much all of the changes here are fixing minor things like that.
* Newest versions of the grunt plugins for JSHint and Karma are being used now, mostly so I could run JSHint over the test spec files.


jni:
@starcalibre I like it! Linters are just amazing. =)

